Issue #7, opened 2017-10-10
Add MacOS support


juanjux:
The native libuast shared library must be built (or downloaded) for MacOS.

juanjux:
Fixed by #22

PR #13, opened 2017-10-10
Add libuast bindings

juanjux:


README.md
```diff
@@ -3,8 +3,54 @@
 This a pure Scala implementation of [Babelfish](https://doc.bblf.sh/) client.
 It uses [ScalaPB](https://scalapb.github.io/grpc.html) for Protobuf/gRPC code generation.
 
+### Installation
```
abeaumont: It'd be nice to describe how to install it through a package manager too (something like https://github.com/src-d/enry/tree/master/java#usage)

README.md
```diff
@@ -3,8 +3,54 @@
...
+
+```
+./sbt assembly
+```
+
+gRPC/protobuf are re-generate from `src/main/proto` on every `./sbt compile` and
+are stored under `./target/scala-2.11/src_managed/`.
```
abeaumont: After I added `crossPaths := false` to the build the `scala-2.11` part doesn't exist anymore, so it'd be `./target/srcd_manager/`.

README.md
```diff
@@ -3,8 +3,54 @@
...
+```
+
+gRPC/protobuf are re-generate from `src/main/proto` on every `./sbt compile` and
+are stored under `./target/scala-2.11/src_managed/`. 
+
+The jar file and the native module are generated in the `build/` directory. If 
+you move the jar file to some other path, the native (`.so` or `.dylib`) 
+library must be in the same path.
+
+If the build fails because it can find the `jni.h` header file, run it with:
```
abeaumont: s/can/can't/

README.md
```diff
@@ -3,8 +3,54 @@
...
+
+```
+./sbt -java_path /usr/lib/jvm/java-8-openjdk-amd64 assembly
+```
+
+Changing the JDK directory to the one right for your system.
+
+*Note: currently only Linux is supported, check [this
```
abeaumont: `[this issue](https://github.com/bblfsh/client-scala/issues/7)` could just be replaced by `#7` ;)

PR #20, opened 2017-10-12
fix libuast installation on macOS

erizocosmico:
Result of of `os.name` is `Mac OS X`, not `Mac` and thus in macOS it tries to load `libuast.so`

```
scala> System.getProperty("os.name").toLowerCase
res3: String = mac os x
```


abeaumont:
Thanks!

PR #29, opened 2017-10-19
Implement libuast 1.3 interface

juanjux:
- Implemented the new libuast 1.3 methods.
- Tests.
- Follow style guide.
- Reorganized and refactorized code.


src/main/scala/org/bblfsh/client/libuast/jni_utils.c
```diff
@@ -0,0 +1,241 @@
...
+
+  (*env)->ReleaseStringUTFChars(env, jstr, tmp);
+  if ((*env)->ExceptionOccurred(env))
+    return NULL;
+
+  return cstr;
+}
+
+jobject *ToObjectPtr(jobject *object) {
+  jobject *copy = malloc(sizeof(jobject));
```
abeaumont: Who takes care of this memory?

juanjux: It's only used in `ChildAt`; in that case the returned node is owned by libuast (which uses it for a call to `CreateXmlNode` (assigns it to the `xmlNodePtr->private` which is freed by libuast's `xmlFreeNode`).

abeaumont: `xmlFreeNode` doesn't free `private` pointer (it cannot know whether it should), so I'd say we're leaking memory here.

juanjux: Damn, I just saw the comments about "libxml won't touch it" in the libxml2 doc. Let me investigate how to handle this, since it's a `jobject` returned by the JNI api by value.

PR #31, opened 2017-10-19
Fix macOS build.

juanjux:
Signed-off-by: Juanjo Alvarez <juanjo@sourced.tech>


juanjux:
@abeaumont build 1.3.0 (which I'll delete and recreate as this one) failed so probably the staging thing on Maven will fail again; fix that when you can.

I'll check that leak you found after this emergency PR.

PR #36, opened 2017-10-20
Fix for memory leak + memory usage test

### LINKED ISSUE #33: Memory leak at ChildAt implementation
Libxml2's xmlFreeNode won't free the `_private` member so that function implementation in `nodeiface.c` would be leaking a node.

juanjux:
Since the `ChildAt` interface function requires a (`void *`) pointer but the JNI API always returns jobjects by value (to clean them up when they go out of scope), a malloc + copy was needed, also because the `jobject` struct is opaque and thus doesn't provide access to the internal object. 

This probably caused a memory leak, so I've added an object to keep track of dynamic allocations and free them at the end of `filter()`.

Also: a simple memory usage test, a little cleanup of the `build.sbt` file and a bump to version 1.3.3.

Fixes #33


src/test/scala/org/bblf/client/BblfshClientTest.scala
```diff
@@ -128,5 +128,30 @@ class BblfshClientTest extends FunSuite with BeforeAndAfter {
...
+    }
+    someFilter
+    val initialUsage = getMemUsage
+
+    for (a<-1 to 10000) {
+      someFilter
+    }
+
+    val finalUsage = getMemUsage
+    assert(finalUsage <= initialUsage * 1.03)
```
abeaumont: This seems pretty ad-hoc... and will probably make this test unstable.
Using relative values also doesn't help.

src/main/scala/org/bblfsh/client/libuast/alloclist.c
```diff
@@ -0,0 +1,38 @@
...
+extern "C" {
+#endif
+
+#include "alloclist.h"
+
+#include <jni.h>
+
+#include <stdlib.h>
+
+void initAllocList(AllocList *l, size_t initialSize) {
```
abeaumont: Calling this a list when it's actually a vector looks weird.

src/main/scala/org/bblfsh/client/libuast/org_bblfsh_client_libuast_Libuast.c
```diff
@@ -22,6 +26,8 @@ JNIEXPORT jobject JNICALL Java_org_bblfsh_client_libuast_Libuast_filter
   if ((*env)->MonitorEnter(env, self) != JNI_OK)
     goto exit;
 
+  initAllocList(&allocList, ALLOCLIST_SIZE);
```
abeaumont: Minor: Initial vector size is probably better left as an implementation detail.

src/main/scala/org/bblfsh/client/libuast/alloclist.c
```diff
@@ -0,0 +1,38 @@
+#ifdef __cplusplus
```
abeaumont: Minor: I don't think this code is expected to be called from C++, since this is ad-hoc code for JNI.
OTOH, if you're using C++, you already have vectors and you wouldn't need this. Can this be built with a C++ compiler instead?

src/main/scala/org/bblfsh/client/libuast/alloclist.c
```diff
@@ -0,0 +1,38 @@
...
+
+void freeAllocList(AllocList *l) {
+  int i;
+  for (i=0; i < l->used; i++) {
+    free(l->list[i]);
+  }
+
+  free(l->list);
+  l->list = NULL;
+  l->used = l->size = 0;
```
abeaumont: If you call `trackAllocatedJObject` after this, you have memory corruption. A sanity check in `trackAllocationJObject` may be a good idea.

src/main/scala/org/bblfsh/client/libuast/alloclist.c
juanjux: Named it list in the generic sense, but I'll rename it to vector.

src/main/scala/org/bblfsh/client/libuast/alloclist.c
juanjux: I would prefer to keep it plain C for the moment to avoid dragging the C++ std into the .so (complicating the macOS build, etc). If we need more features in the future we could always move to C++.

src/main/scala/org/bblfsh/client/libuast/alloclist.c
juanjux: No corruption, it would read a NULL pointer which would cause a SIGSEGV. Another option would be to detect that the vector is not initialized on `trackAllocatedJObject` and do the initialization there. What do you think?

src/main/scala/org/bblfsh/client/libuast/org_bblfsh_client_libuast_Libuast.c
juanjux: :+1:

src/test/scala/org/bblf/client/BblfshClientTest.scala
juanjux: Yeah, it's imprecise which unittests shouldn't be; but it runs in the context of the unittests which is somewhat of a controlled environment. I'll remove it anyway.

src/main/scala/org/bblfsh/client/libuast/alloclist.c
juanjux: PS: check the last commit, I just did that (removed `initAllocVector` from the public interface and made `trackAllocatedObject` call it if needed).

abeaumont: No, it would not read a NULL pointer. It would check that `size` is equal to `used` and then would call `realloc` with NULL and 0 `size`, which may or may not return NULL, then that pointer will be accessed... so it doesn't sound safe at all :)

This should be fixed with your last commit though (but I wouldn't make the vector struct public, though.)

src/main/scala/org/bblfsh/client/libuast/alloclist.c
abeaumont: I'd say having a custom vector implementation is a good enough reason, but I don't know how much it complicates the build. Anyway, going back to my original question, if we're staying in plain C, what do we need to add these C++ compatibility checks? Not that it's important, just curious.

src/main/scala/org/bblfsh/client/libuast/alloclist.c
juanjux: I checked that realloc with the NULL argument produces a sigsev in that case.

Anyway... it doesn't work :-| Since the freeing is at the end of `filter()` call and `filter` returns data, freeing part of that data produces a crash later if you try to use it (from Scala), so I'll take a look at how to hook with JNI into the Scala object destructor to call the free there.

src/main/scala/org/bblfsh/client/libuast/alloclist.c
juanjux: It's pretty common to do that in plan C; if somebody compiles the file with a C++ compiler it will wrap everything inside an "extern C" so the compiler will threat the code inside as "plain old C" (C and C++ nowadays are not 100% compatible). It's not important anyway since we specify the compiler in the build, maybe I'll remove it.

abeaumont: Yes, I know what is for, that's why I asked for this particular usage, since it's plain C, not expected to be used elsewhere, and we control the compiler (that's why I said that if we're expecting to use C++ we wouldn't need this custom code in the first place ;)

juanjux: After my debugging session tracking pointers to pointers I'll probably move it to C++ and std:vector sooner than later hehe. I've not done it already because JNI has a different (better) API for C++ and I would like to migrate to it too, but it's not trivial.

src/main/scala/org/bblfsh/client/libuast/alloclist.c
juanjux: Luckily the problem wasn't that complex but an `extern` declaration of the vector as a pointer (extern AllocVector *allocVector)with the real declaration as non pointer... surprising that `gcc` didn't even issue a warning, but such is life. That was fixed following your advice of making the vector private to the file :+1:

src/main/scala/org/bblfsh/client/libuast/alloclist.c
juanjux: PS: I also re-added the synchonization in the Scala part now that we've a working way to do it (with the `Libuast` object but outside of it). Works well under spark, BTW it would be nice to test from Spark in Travis; I need to check how to run jupiter notebooks headless and programatically.

src/main/scala/org/bblfsh/client/libuast/objtrack.c
```diff
@@ -0,0 +1,60 @@
...
+                           _allocVector->size * sizeof(jobject*));
+  }
+  _allocVector->vector[_allocVector->used++] = obj;
+}
+
+void freeObjects() {
+  if (_allocVector == NULL)
+    return; // No object ever tracked thus not initialized
+
+  int i;
```
abeaumont: Why is this defined outside the loop context?

juanjux: It was to avoid warnings with Travis' Ubuntu 14.04 GCC about C99 features, but since I added the C99 switch is should be unnecessary, I'll change it.

PR #44, opened 2017-11-02
Make Node.filter() work.

### LINKED ISSUE #37: Make Node.filter() work
It would be more natural and Scala'ish.

Probably can be implemented using a Scala's `implicit` object or method taking Nodes.

juanjux:
Fixed #37.

- Added implicit class in the Object for calling Node.filter.

- Moved filter to the Object (static Scala style) since it doesn't need
  the this pointer and it's needed there for the implicit object to work
  without a reference to the client instance.

- Reworked test to use Node.filter and added a shortcut to resp.uast.get.

Signed-off-by: Juanjo Alvarez <juanjo@sourced.tech>

### DIFF
--- a/src/main/scala/org/bblfsh/client/BblfshClient.scala
+++ b/src/main/scala/org/bblfsh/client/BblfshClient.scala
@@ -9,7 +9,7 @@ import gopkg.in.bblfsh.sdk.v1.uast.generated.Node
 import io.grpc.ManagedChannelBuilder
 
 
-class BblfshClient(host: String, port: Int, maxMsgSize: Int) { 
+class BblfshClient(host: String, port: Int, maxMsgSize: Int) {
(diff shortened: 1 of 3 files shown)


src/main/scala/org/bblfsh/client/BblfshClient.scala
```diff
@@ -30,13 +29,15 @@ class BblfshClient(host: String, port: Int, maxMsgSize: Int) {
     parsed
   }
 
-  def filter(node: Node, query: String): List[Node] = Libuast.synchronized {
-    libuast.filter(node, query)
+  // proxy to the static method for backward compatibility
```
abeaumont: Maybe better to use [scaladoc style](https://docs.scala-lang.org/style/scaladoc.html)?

PR #46, opened 2017-11-02
Support native parse and version methods

### LINKED ISSUE #45: Add support for new protocol methods NativeParse and Version

abeaumont:
Fixes #45

### DIFF
--- a/src/main/scala/org/bblfsh/client/BblfshClient.scala
+++ b/src/main/scala/org/bblfsh/client/BblfshClient.scala
@@ -2,8 +2,10 @@ package org.bblfsh.client
 
 import org.bblfsh.client.libuast.Libuast
 
-import gopkg.in.bblfsh.sdk.v1.protocol.generated.{Encoding, ParseRequest, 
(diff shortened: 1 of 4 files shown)


src/main/scala/org/bblfsh/client/BblfshClient.scala
```diff
@@ -25,8 +27,23 @@ class BblfshClient(host: String, port: Int, maxMsgSize: Int) {
                            content = content,
                            language = BblfshClient.normalizeLanguage(lang),
                            encoding = encoding)
-    val parsed = stub.parse(req)
-    parsed
+    stub.parse(req)
+  }
+
+  def parseNative(name: String, content: String, lang: String = "",
```
juanjux: `parseNative` or `nativeParse`? We're using "native-first" naming in the SDK and protocol (I was using parse_native for Python).

abeaumont: oh, my bad, fixed now.

PR #57, opened 2018-01-14
Support for LibUAST 1.6.1 iterators + other improvements.

juanjux:
- Adds support for libuast iterators.
- Partial migration to C++.
- New memory tracker to avoid issues with JVM memory rellocation of objects between JNI calls and reference counting.

Depends on bblfsh/libuast/pull/64 so CI will fail until that is merged and libuast 1.7.0 is released.


bzz:
@juanjux very cool! Let me look into this later today or early tomorrow.

juanjux:
@bzz thanks, feel free to add a review to the code if you want too. I'll probably add a `node.iterate` method in the future to be more idiomatic and in line with `node.filter`.

Issue #68, opened 2018-04-10
Add support to query supported languages (with installed drivers)


smola:
In some applications, it is very useful to know which languages are supported (with installed drivers) beforehand. This helps to:

* Display lists of supported languages in UX elements.
* Reduce number of requests to bblfshd by avoiding requests for unsupported languages.

It would be nice to have such functionality implemented in the Scala client.

dpordomingo:
We have that necessity for https://github.com/bblfsh/dashboard/issues/89 too.
It [was proposed](https://github.com/bblfsh/dashboard/issues/89#issuecomment-351655057) to use the `github.com/bblfsh/bblfshd/daemon/protocol.ProtocolServiceClient`, but it currently runs over a unix socket so it is not reachable from the outside of the bblfsh container.

bzz:
Indeed we need this, but for golang client first :)

This makes me think - do you guys think this is the right repo to track this, or is there a better one if changes are required on a protocol-level?

juanjux:
Closing since we just got a duplicated in the SDK

sdk/issues/253

juanjux:
Still needs to be integrated in the client, reopening.

PR #69, opened 2018-06-15
Add support to query supported languages

smacker:
Fix: #68

1. Update sdk version
2. Add `supportedLanguages` method and test
3. Add `"-I" + javaHome + "/include/darwin "` to libuast build.sbt. It allows to build `client-scala` on mac.

### DIFF
--- a/src/test/scala/org/bblf/client/BblfshClientSupportedLanguagesTest.scala
+++ b/src/test/scala/org/bblf/client/BblfshClientSupportedLanguagesTest.scala
@@ -12,7 +12,11 @@ class BblfshClientSupportedLanguagesTest extends FunSuite with BeforeAndAfter {
     resp = client.supportedLanguages()
   }
 
-  test("Check languages") {
+  test("Check languages are not empty") {
     assert(!resp.languages.isEmpty)
+  }
+
+  test("Check languages contain Java") {
+    assert(resp.languages.map(d => d.name).contains("Java"))
   }
 }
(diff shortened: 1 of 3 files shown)


src/test/scala/org/bblf/client/BblfshClientSupportedLanguagesTest.scala
```diff
@@ -0,0 +1,18 @@
...
+class BblfshClientSupportedLanguagesTest extends FunSuite with BeforeAndAfter {
+  val client = BblfshClient("0.0.0.0", 9432)
+  var resp: SupportedLanguagesResponse = _
+
+  before {
+    resp = client.supportedLanguages()
+  }
+
+  test("Check languages") {
+    assert(!resp.languages.isEmpty)
```
juanjux: This could fail if the bblfshd didn't have any drivers installed while not really being a failure in the client. Could you add an exec() to run the driver install command so we are sure there is at least a driver installed and then check that resp.languages contain that specific language?

smacker:
@juanjux 
Could you please point me where do you want to add `exec`?

There is a command in `.travis.yml` already: `docker exec -it bblfsh bblfshctl driver install --all`

Other tests rely on this as well. Both `ClientNativeParseTest` and `ClientParseTest` assume java driver was installed and will fail otherwise.

Originally I was following the check in [client-go](https://github.com/bblfsh/client-go/blob/master/client_test.go#L66). Now I added check for `Java`. Is it good enough?

juanjux:
Yes, I forgot that the CI adds the Java driver, still not perfect because local tests could fail if the user doesn't have the driver installed but that is not a problem of this PR.

Issue #71, opened 2018-06-22
Expose method for gRPC connection termination


bzz:
Right now `BblfshClient` manages the connection creation, but it relies on GC to shut the connection down.

It would be nice to expose something gRCP's channel `.shutdownNow()` method to allow clients to for more fine-grained resource control and to avoid this message on a process termination (i.e in tests):

```
ERROR 00:53:24,580 io.grpc.internal.ManagedChannelImpl$ManagedChannelReference (ManagedChannelImpl.java:1236) - *~*~*~ Channel io.grpc.internal.ManagedChannelImpl-1 for target 127.0.0.1:9432 was not shutdown properly!!! ~*~*~*
    Make sure to call shutdown()/shutdownNow() and awaitTermination().
java.lang.RuntimeException: ManagedChannel allocation site
	at io.grpc.internal.ManagedChannelImpl$ManagedChannelReference.<init>(ManagedChannelImpl.java:1185)
	at io.grpc.internal.ManagedChannelImpl.<init>(ManagedChannelImpl.java:504)
	at io.grpc.internal.AbstractManagedChannelImplBuilder.build(AbstractManagedChannelImplBuilder.java:369)
	at org.bblfsh.client.BblfshClient.<init>(BblfshClient.scala:19)
	at org.bblfsh.client.BblfshClient$.apply(BblfshClient.scala:81)
```

Will be happy to provide a patch, in case there is a consensus of usefulness of such method.

bzz:
@dennwc @juanjux what do you guys think?

juanjux:
I agree that it would be useful and patches are welcome, if not I could do it as soon as I finish my current task (tomorrow probably).

PR #72, opened 2018-06-29
Expose method for gRPC connection termination

smacker:
Fix: #71

### DIFF
--- a/src/main/scala/org/bblfsh/client/BblfshClient.scala
+++ b/src/main/scala/org/bblfsh/client/BblfshClient.scala
@@ -52,6 +52,10 @@ class BblfshClient(host: String, port: Int, maxMsgSize: Int) {
     stub.version(req)
   }
 
+  def close(): Unit = {
+    channel.shutdownNow()
+  }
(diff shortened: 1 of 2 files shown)


src/main/scala/org/bblfsh/client/BblfshClient.scala
```diff
@@ -52,6 +52,10 @@ class BblfshClient(host: String, port: Int, maxMsgSize: Int) {
     stub.version(req)
   }
 
+  def shutdownNow(): Unit = {
```
smola: maybe expose as the more standard `close()`?

smacker: the name is taken from the issue description. I don't have any opinion about which name is better. Maybe maintainer @juanjux should decide?

juanjux: I agree, close() is better (same as files sockets, etc)

smacker:
renamed & amended & repushed

PR #82, opened 2019-05-14
Initial build fix-up

bzz:
Housekeeping, consisting of mostly small, cosmetic changes:
 - make default build pass on macOS
 - centralize all SBT plugin configuration in a single file (instead of may small)
 - remove whitespaces in main doc
 - fix test that will fail locally, as bblfsh is too fast now 🎸 
 - bump SDK/libuast versions to latest compatible
 - (new) fix native lib loading so tests actually work in IDE as well

### DIFF
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -1,0 +1,6 @@
+addSbtPlugin("com.eed3si9n" % "sbt-assembly" % "0.14.5")
+addSbtPlugin("net.virtual-void" % "sbt-dependency-graph" % "0.8.2")
+addSbtPlugin("com.jsuereth" % "sbt-pgp" % "1.0.0")
+addSbtPlugin("org.xerial.sbt" % "sbt-sonatype" % "1.1")
+addSbtPlugin("com.thesamet" % "sbt-protoc" % "0.99.16")
+libraryDependencies += "com.thesamet.scalapb" %% "compilerplugin" % "0.8.3"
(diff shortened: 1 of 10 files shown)


src/test/scala/org/bblfsh/client/BblfshClientTimeoutTest.scala
```diff
@@ -10,8 +10,8 @@ class BblfshClientTimeoutTest extends FunSuite with BeforeAndAfter {
   val fileName = "src/test/resources/large.php"
   val fileContent = Source.fromFile(fileName).getLines.mkString("\n")
 
-  test("Parse requests with default timeout fails") {
-    val resp = client.parse(fileName, fileContent)
+  test("Parse requests with small timeout fails") {
+    val resp = client.parseWithTimeout(fileName, fileContent, 1)
```
creachadair:
```suggestion
    val resp = client.parseWithTimeout(fileName, fileContent, 1 /* seconds */)
```

bzz:
@creachadair feedback addressed, ready for another round.

PR description updated, scope now also include fix for a built so tests can be run from IDE as well!

project/plugins.sbt
```diff
@@ -0,0 +1,6 @@
+addSbtPlugin("com.eed3si9n" % "sbt-assembly" % "0.14.5")
+addSbtPlugin("net.virtual-void" % "sbt-dependency-graph" % "0.8.2")
+addSbtPlugin("com.jsuereth" % "sbt-pgp" % "1.0.0")
+addSbtPlugin("org.xerial.sbt" % "sbt-sonatype" % "1.1")
+addSbtPlugin("com.thesamet" % "sbt-protoc" % "0.99.16")
+libraryDependencies += "com.thesamet.scalapb" %% "compilerplugin" % "0.8.3"
```
creachadair: I'm a little fuzzy on how the dependency resolution works here, but would we not want `%%` for the other plugins as well? (Or doesn't the Scala version matter there?)

bzz: That's a good question - AFAIK there is an option of passing specific `scalaVersion` to the `addSbtPlugin` but we usually do not use it, thus allowing for SBT to choose a version of Scala that SBT itself was built across, rather then the one we are building for.

bzz:
Merging if there is no further discussion

Issue #83, opened 2019-05-20
Migrate scala-client to SDK.v3 + libuast.v3


bzz:
An umbrella issue for client migration to new SDK. 

Takes similar approach to bblfsh/client-python. but aims to avoid having [all the changes in a single PR](https://github.com/bblfsh/client-python/pull/128), so will be broken down.

 - [x] SDK.v2 - parse API
   - [x] `parse()` impl, that uses new `protocol.v2`
   - [x] `decode()` impl, so the binary wire protocol can be read (on Go side)
   - [x] `root()` impl, so the decoded (on Go side) UAST can be exposed to JVM
   - [x] `load()` impl, so the Go-backed UAST can be transferred to host JVM memory #91 
   - [x] `encode()` impl. Reverse: JVM memory objects transferred to Go 
   - [x] expose mode argument for parse (native, annotated, semantic) #111
 - [x] SDK.v2 - filter API
   - [x] `iterator()` impl, so UAST can be iterated with and without loading it to JVM #106
   - [x] `filter()` impl, so XPath filtering works #109 
            take [early feedback on API](https://reviewable.io/reviews/bblfsh/scala-client/97#-LjMH6RtA3Mz0RtfWryx) into account 
   - [x] `Context` impl (JVM-side factory), so multiple concurrent `filter()` calls from the client are supported 
 - [x] SDK.v2 - list languages API
    Support for languages/version request (as soon as https://github.com/bblfsh/sdk/pull/400 is released)
 - [x] doc: add JNI debugging instructions #115 
 - [x] doc: update API examples in README and CLI tool
 - [x] ~investigate JNI warnings~ moved to #112 
 - [x] ~enable timeouts at v2 API again~ moved to #100
 - [x] ~CI: refactor to multiple modular jobs (tests, releases per-platform, etc)~ moved to #87

bzz:
### New API example

Protocol.v2 response is protobuf-like encoded binary blob, so libuast now provides a method to `decode()` it on the Go side. Decoding will result only in in-memory UAST representation on the Go side, thus in order to be exposed to the host environment (JVM) as objects - the root node is also needs to `load()`ed.

Examples below use explicit type excessively in order to highlight the API internals, real applications \w same code will be more concise.

#### decode + load (Go->JVM)
```scala
import org.bblfsh.client.v2.*

// parse
//// read the binary UASTv2
val resp = client.parse(fileName, fileContent)

//// decode it, UAST is backed by Go
[...]

PR #84, opened 2019-05-28
Migrate to SDK.v3 - protocol.v2 with decode

bzz:
Partially addresses #83 

 - [x] new API is under `org.bblfsh.client.v2`
 - [x] new SDK and libraste is used
 - [x] `decode()` and `root()` are implemented
 - [x] add CI profile only with things that are expected to pass at current stage
 - [x] add 2 TODOs base on feedback
 - [x] enable test for the latest bblfshd/sdk release
 - [x] fix CI v2 profile
   Right now failing with `java.lang.IllegalStateException: Linux build failed -> fatal error: libuast.h: No such file or directory`
 - [x]  API visibility in native lib
 - [x] ~A bit improved tests of `.decode()`~ 
   As C++ API does not expose `.Kind()` for nodes, this can be done later, after `.load()` is implemented.

Main CI is expected to fail due to still missing native API that will be added in separate PRs.


src/main/scala/org/bblfsh/client/v2/BblfshClient.scala
```diff
@@ -0,0 +1,161 @@
...
+    *                Right now this does not have any effect in v2.
+    * @return UAST in parse response.
+    */
+  def parseWithOptions(
+    name: String,
+    content: String,
+    lang: String,
+    timeout: Long
+  ): ParseResponse = {
+    // FIXME(bzz): make timout work in v2 again
```
dennwc: Oh, I see, so it's gone from the request now and should be set by other means. I guess it's good for the client since now it will also know that we have a timeout for the RPC call.

src/main/scala/org/bblfsh/client/v2/BblfshClient.scala
```diff
@@ -0,0 +1,161 @@
...
+    * @param name file name
+    * @param content file content
+    * @param lang (optional) language to parse, default auto-detect \w enry
+    * @return UAST in parse response.
+    */
+  def parse(
+    name: String,
+    content: String,
+    lang: String = ""
+  ): ParseResponse = parseWithOptions(name, content, lang, DEFAULT_TIMEOUT_SEC)
```
dennwc: Maybe we can just add another optional parameter for the timeout? It shouldn't break the caller, right?

src/main/scala/org/bblfsh/client/v2/BblfshClient.scala
```diff
@@ -0,0 +1,161 @@
...
+
+  /**
+   * Decodes bytes from wired format of bblfsh protocol.v2.
+   * Requires a buffer in Direct mode.
+   *
+   * Since v2.
+   */
+  def decode(buf: ByteBuffer): Context = Libuast.synchronized {
+    if (!buf.isDirect()) {
+      throw new RuntimeException("Only directly-allocated buffer decoding is supported.")
```
dennwc: Is it possible to convert it to a direct buffer here? It will probably cost allocation, but it will work at least. Or is that a rare use case?

src/main/scala/org/bblfsh/client/v2/Node.scala
```diff
@@ -0,0 +1,5 @@
+package org.bblfsh.client.v2
+	
+case class Node(ctx: Long, handle: Long) {
+  @native def load(): Map[String, _]
```
dennwc: I'm not sure it will return `Map[string, _]` in all cases. The node may be a single string value or an array, for example.

src/test/scala/org/bblfsh/client/v2/BblfshClientClose.scala
```diff
@@ -1,13 +1,12 @@
...
 
-import gopkg.in.bblfsh.sdk.v1.protocol.generated.VersionResponse
-import org.scalatest.FunSuite
-import org.scalatest.BeforeAndAfter
+import gopkg.in.bblfsh.sdk.v2.protocol.driver.VersionResponse
+import org.scalatest.{BeforeAndAfter, FunSuite, Ignore}
 
 import scala.io.Source
 
+@Ignore // unitll we have a bblfshd release supporting aliases
```
dennwc: It's now released :)

src/test/scala/org/bblfsh/client/v2/BblfshClientParseTest.scala
```diff
@@ -0,0 +1,71 @@
...
+    println(uast)
+    uast.dispose()
+  }
+
+  "Decoded UAST RootNode" should "not be NULL" in {
+    val uast = resp.uast.decode()
+    val rootNode: Node = uast.root()
+    println(rootNode.getClass)
+
+    rootNode should not be null
```
dennwc: Can we also assert the structure, at least for the root node? That it has specific keys with specific types, for example. Checking the whole tree is even better :wink:

src/main/scala/org/bblfsh/client/cli/ScalaClientCLI.scala
```diff
@@ -24,12 +25,12 @@ object ScalaClientCLI extends App {
   if (resp.errors.isEmpty) {
     if (query != null && query != "") {
       try {
-        println(BblfshClient.filter(resp.uast.get, query))
+        println(BblfshClient.filter(resp.uast.decode().root(), query))
```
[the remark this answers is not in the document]
bzz: AFAIK it does not matter, both would do the same

src/main/scala/org/bblfsh/client/cli/ScalaClientCLI.scala
```diff
@@ -1,6 +1,7 @@
 package org.bblfsh.client.cli
 
-import org.bblfsh.client.BblfshClient
+import org.bblfsh.client.v2.BblfshClient
+import org.bblfsh.client.v2.BblfshClient._
```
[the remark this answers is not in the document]
bzz: It's needed in order for implicits defined in there to work, so `resp.uast.decode` becomes a thing.

src/main/scala/org/bblfsh/client/v2/BblfshClient.scala
bzz: Not sure what you mean, for now it just keeps the same API as it was before.

src/main/scala/org/bblfsh/client/v2/BblfshClient.scala
bzz: This is a low-level API with documented preconditions on arguments.
A convenience wrapper that does the conversions is [defined below](https://github.com/bblfsh/client-scala/pull/84/files#diff-125e1643be081078c4dc9eec085a2517R141)

src/main/scala/org/bblfsh/client/v2/Node.scala
bzz: Good catch, thank you, will add a comment not to forget when working on `local()`

src/test/scala/org/bblfsh/client/v2/BblfshClientParseTest.scala
bzz: Sure, we will improve tests as we move on.

BTW, could you help me understand what do you mean by `assert the structure` here, without th `load()` or any iterators?

src/main/scala/org/bblfsh/client/v2/BblfshClient.scala
dennwc: I mean that there is a `parseWithOptions` function that seems unnecessary. If I understood correctly, the `parse` function may include the `timeout` argument with a default value, so there is only one call instead of two, but it's still backward-compatible because the parameter is optional. I'm probably missing something here.

src/main/scala/org/bblfsh/client/cli/ScalaClientCLI.scala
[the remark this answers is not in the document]
dennwc: Can you please help me understand, this method attaches to any `ByteBuffer` as long as the import is enabled, right? Won't it collide with `decode` defined by some other package?

src/test/scala/org/bblfsh/client/v2/BblfshClientParseTest.scala
dennwc: Yep, sure. The Nodes API works symmetrically for both client-hosted and `libuast`-hosted nodes, so it's possible to call methods like `Kind` to assert a type of a UAST node or get the list of keys the same way. You are right, it doesn't make sense to go deep before the `load` is implemented, but asserting few basic things might help to catch some bugs that may be triggered before the `load`.

src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc
```diff
@@ -0,0 +1,136 @@
+#include "org_bblfsh_client_v2_libuast_Libuast.h"
+#include "jni_handle.h"
+#include "jni_utils.h"
+#include "org_bblfsh_client_v2_Context.h"
+
+#include "libuast.h"
+#include "libuast.hpp"
+
+JavaVM *jvm;  // FIXME(bzz): double-check and document
```
creachadair: We should almost certainly put this symbol into a namespace unless for some reason we can't. I'm also concerned about initialization, since there is a link-time dependency with `jni_utils.cc` and I can't currently tell for sure that we actually get the correct order. 

I second your TODO, and would add that we should document not only what it is, but how and where it is populated. If you haven't already considered it, I would think about providing a factory function that does the initialization on demand and returns a pointer, which would allow us to anonymously namespace the symbol and protect us from dependency order oddities up to concurrency.

```c
JavaVM* getJvm() {
    static JavaVM* instance = nullptr;
    static bool initialized = false;

    if (!initialized) {
        instance = new JavaVM;  // or whatever
        initialized = true;
    }
    return instance;
}
```

or words to that effect. (Or in an anonymous namespace at file scope, if you dislike statics)

src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc
```diff
@@ -0,0 +1,136 @@
...
+
+jobject asJvmBuffer(uast::Buffer buf) {
```
creachadair: Here and throughout, please consider wrapping utility code required only by the current translation unit in anonymous namespace(s).

src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc
```diff
@@ -0,0 +1,136 @@
...
+JNIEXPORT void JNICALL Java_org_bblfsh_client_v2_Context_dispose(JNIEnv *env,
+                                                                 jobject self) {
+  ContextExt *p = getHandle<ContextExt>(env, self, "nativeContext");
+  setHandle<ContextExt>(env, self, 0, "nativeContext");
+  delete p;
+}
+
+JNIEXPORT jobject JNICALL Java_org_bblfsh_client_v2_libuast_Libuast_filter(
+    JNIEnv *, jobject, jobject, jstring) {
+  return NULL;  // TODO(bzz): implement
```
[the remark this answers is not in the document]
creachadair: As far as I'm aware, the JNI machinery does not support C++ exceptions. In principle we could catch and convert them with a wrapper, but inasmuch as we can avoid using C++ exceptions at all I think we'll be better off. The closer we can keep these stubs to plain C I believe the better off we are.

src/main/scala/org/bblfsh/client/v2/BblfshClient.scala
```diff
@@ -0,0 +1,161 @@
...
+import com.google.protobuf.ByteString
+import com.google.protobuf.duration.Duration
+import io.grpc.ManagedChannelBuilder
+import gopkg.in.bblfsh.sdk.v2.protocol.driver.{ParseRequest, ParseResponse, DriverGrpc, SupportedLanguagesRequest, SupportedLanguagesResponse, VersionRequest, VersionResponse, DriverHostGrpc}
+
+import org.bblfsh.client.v2.libuast.Libuast
+
+
+class BblfshClient(host: String, port: Int, maxMsgSize: Int) {
+  private val DEFAULT_TIMEOUT_SEC = 5
```
creachadair: Do we have to have a default? And if so, can we make it fairly long? As we've seen, it's hard to debug weird timeout failures deep inside the plumbing. Assuming this is only here as a safety backstop so a bug doesn't wedge us forever, it should be fine to have it be like 1m or 2m. "This always fails after around 1 minute" is easy to diagnose vs. "This often but not always times out around 5 seconds".

src/main/scala/org/bblfsh/client/v2/BblfshClient.scala
```diff
@@ -0,0 +1,161 @@
...
+    stubInfo.serverVersion(req)
+  }
+
+  def close(): Unit = {
+    channel.shutdownNow()
+  }
+}
+
+object BblfshClient {
+  val DEFAULT_MAX_MSG_SIZE = 100 * 1024 * 1024
```
creachadair:
```suggestion
  val DEFAULT_MAX_MSG_SIZE = 100 * 1024 * 1024  // bytes
```

creachadair:
> First round is done! I don't understand most of the JNI code, but at least the Scala code is clear. Not marked as approved because there are still few changes needed (like mentioned in the `FIXME`).

It's funny, because I was just feeling much more comfortable with the C++ portion. Which I suppose says something about my poor life-choices. 😀

build.sbt
```diff
@@ -149,29 +148,30 @@ def compileUnix(sourceFiles: String) = {
...
 
     checkedProcess(cmd, "macOS build")
   } else {
     val cmd:String = "g++ -shared -Wall -fPIC -O2 -std=c++11 " +
       "-I/usr/include " +
       "-I" + javaHome + "/include/ " +
       "-I" + javaHome + "/include/linux " +
-      "-Isrc/libuast-native/  " +
+      "-Isrc/main/resources/libuast " +
+      "-Lsrc/main/resources/libuast " +
```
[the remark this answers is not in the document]
bzz: I know, but the reason is that the libraries have to be in the class path of the test runner, in order for test to run in both, CLI and IDE. And I think it's ok, as soon as they are in resources

src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc
bzz: This is a very valuable feedback - thank you! That's a piece of code left from previous implementation, it's now initialized while loading the JNI (guaranteed to be once) and if you do not mind, I would prefer dealing with this in a subsequent PRs, as for now it's "just the same as before" and should not be treated as a change. 

I'll add a TODO with the link to this thread but please, let me know if you think it's crucial to have some immediate improvements (e.g. namespace)

src/main/scala/org/bblfsh/client/v2/BblfshClient.scala
bzz: I know that Github does not make it simpler to distinguish, but this is not a change but rather the same code as before. There is a TODO for dealing with timeouts, both in code and an the umbrella issue description, so I've attached the link in there of feedback not to get lost.

I appreciate the feedback, but if possible, would prefer focusing this and few subsequent PRs on functionality related to SDK/libuast move.

src/main/scala/org/bblfsh/client/v2/BblfshClient.scala
bzz: Thanks! Added to timeout todo item in the umbrella issue - if you do not mind, I would prefer to apply sam logic as in https://github.com/bblfsh/client-scala/pull/84#discussion_r288460731 applied

src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc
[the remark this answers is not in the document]
bzz: AFAIK for any JVM implementation JNI can be written in either C or C++ but the support for native exceptions is implementations-specific (e.g may be available on DalivkVM on android but not HostSpot and vice versa).

At this point, it really does not matter as it's not implemented :) 

But JFY, [previous implementation](https://github.com/bblfsh/client-scala/blob/master/src/main/scala/org/bblfsh/client/libuast/org_bblfsh_client_libuast_Libuast.cc#L177) uses exceptions, but I'm not exactly sure why exactly the same discussion did not happen at the review time, when those changes were added.

src/main/scala/org/bblfsh/client/cli/ScalaClientCLI.scala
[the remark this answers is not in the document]
bzz: >  this method attaches to any ByteBuffer as long as the import is enabled, right
right

>  Won't it collide with decode defined by some other package?

if some other package adds the same implicit, that gets imported here as well?
It will, with compile-time error.

src/main/scala/org/bblfsh/client/v2/BblfshClient.scala
bzz: Yup, and to clarify the nature of this change as well as migration part for the clients as well as adjust up the API - there is a separate todo item in the umbrella issue.

src/main/scala/org/bblfsh/client/v2/BblfshClient.scala
creachadair: > [I]f possible, would prefer focusing this and few subsequent PRs on functionality related to SDK/libuast move.

Yes, that makes sense to me. And as you observed, I simply did not notice that this was the same as the previous behaviour.

src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc
creachadair: This can definitely wait.

src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc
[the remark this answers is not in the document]
creachadair: Consistency is good in a shared interface, but I'd argue that JNI wrappers are morally equivalent to generated code (even though in this case we wrote them by hand) to pave over interoperability issues. 

If we have a reason to make them explicitly consistent, I'd argue we should think about pulling out a library to share between the clients—but in this case I don't think there's any real maintenance gain by trying to make them superficially similar.

src/main/native/jni_handle.h
```diff
@@ -0,0 +1,34 @@
...
+  }
+
+  jfieldID jfid = env->GetFieldID(cls, name, "J");
+  if (env->ExceptionOccurred() || !jfid) {
+    return nullptr;
+  }
+  return jfid;
+}
+
+template <typename T>
```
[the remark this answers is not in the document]
bzz: > Does this need to be a template? 

Sorry for confusion, I belive it does, as later on we'll have at least `getHandle<ContextExt>` and `getHandle<NodeHandle>`

> could we put the definitions out-of-line

could you please give an example, to make sure I got this one right?

src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc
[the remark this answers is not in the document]
bzz: Agreed, let's think about such refactoring possibility after initial versions is running.

src/main/scala/org/bblfsh/client/cli/ScalaClientCLI.scala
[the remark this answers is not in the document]
bzz: Resolving, as presume this answers your question @dennwc . But please, feel free to un-resolve it if it's not the case.

src/main/scala/org/bblfsh/client/v2/Node.scala
bzz: This also means that we can not use https://github.com/bblfsh/client-scala/issues/83#issuecomment-494245514 as we are forced to have some equivalent of `Node` and `NodeExt` distinction, to represent 1) something in JVM that can be `value, array or map`  and  2) a mere proxy to native node.

bzz:
CI is expected to have 1 profile that is passing now. Initial feedback addressed.

2 things that are left:
 - API visibility improvement on the native part 
 - a bit more sophisticated tests for `decode()`

src/test/scala/org/bblfsh/client/v2/BblfshClientParseTest.scala
bzz: > asserting few basic things ...  call methods like Kind to assert a type of a UAST node or get the list of keys

@dennwc would you also be so kind to show me equivalent of this in https://github.com/bblfsh/client-python/ ? Thanks!

src/main/native/jni_handle.h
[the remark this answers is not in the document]
creachadair: I think you're right—it's just that in this PR I only saw `getHandle<ContextExt>`. 

Re: Example. I just meant moving the template definitions to the `.cc` file where they are used, and wrapping them in an anonymous namespace. But if we are later going to use them in multiple `.cc` files that advice should be ignored. 🙂

src/main/scala/org/bblfsh/client/cli/ScalaClientCLI.scala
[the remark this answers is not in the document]
dennwc: I'm okay with it as long as it's a common practice in Scala :)

src/test/scala/org/bblfsh/client/v2/BblfshClientParseTest.scala
dennwc: No. Because it seems there isn't :sweat_smile: 

Opened a https://github.com/bblfsh/client-python/issues/159 to track this. Thanks for checking it!

src/main/native/jni_handle.h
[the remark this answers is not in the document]
bzz: Thank you! For now, we are having most things implemented in a single `.cc` so I would do just that.

src/test/scala/org/bblfsh/client/v2/BblfshClientParseTest.scala
bzz: > The Nodes API works symmetrically for both client-hosted and libuast-hosted nodes, so it's possible to call methods like Kind to assert a type of a UAST node 

@dennwc Could you please point me to an example, how this might have looked in Python API?
On my reading of CPython extension - [PyNodeExt type](https://github.com/bblfsh/client-python/blob/master/bblfsh/pyuast.cc#L48) does not expose "kind", or am I looking at the wrong place?

dennwc: There is definitely a C API for this, but I remember now that the kind was not exposed in C++ API due to some issues with C++ templates.

So it seems like a bigger issue, let's address it in the following PRs.

src/main/scala/org/bblfsh/client/v2/Node.scala
dennwc: I don't think it's practical to allow to load maps only. We'll need to find a way to make it work for other types as well.

bzz: Agreed (was just exploring all possible options).
Will introduce a proper Node abstraction (and something like NodeExt separation) in next PR.

bzz:
@creachadair @dennwc all feedback was addressed, CI v2-enabled profile is green and it's ready to :shipit:

.clang-format
```diff
@@ -0,0 +1,2 @@
+BasedOnStyle: Google
+AllowShortIfStatementsOnASingleLine: true
```
creachadair: I'm surprised this isn't already enabled in the Google style. (Maybe the internal config is—dare I say it—Google _plus_?)

Ahem.

Anyway, though, if you could please add a newline at EOF I'd be grateful. 😁

src/main/scala/org/bblfsh/client/v2/BblfshClient.scala
```diff
@@ -0,0 +1,161 @@
...
+
+  def close(): Unit = {
+    channel.shutdownNow()
+  }
+}
+
+object BblfshClient {
+  val DEFAULT_MAX_MSG_SIZE = 100 * 1024 * 1024 // bytes
+
+  val PreOrder      = 0
```
creachadair: These don't seem to be used anywhere—are they mean to support the iterator implementation that is coming later?

src/main/native/jni_utils.h
```diff
@@ -0,0 +1,13 @@
+#ifndef _Included_org_bblfsh_client_libuast_Libuast_jni_utils
+#define _Included_org_bblfsh_client_libuast_Libuast_jni_utils
+
+#include <jni.h>
+
+extern const char *CLS_NODE;
+extern const char *CLS_CTX;
+
+JNIEnv *getJNIEnv();
+jobject NewJavaObject(JNIEnv *, const char *, const char *, ...);
```
creachadair: I'm not sure I follow the naming convention in use here. [Google style](https://google.github.io/styleguide/cppguide.html#Function_Names) is `MixedCaseLeadingCaps` for classes and functions (with a carve-out for class accessors to follow `snake_case` style). But we seem to be mixing that and conventional camel.

(I'm not wedded to Google style, I only mentioned that because you included a clang-format config based on that style)

build.sbt
```diff
@@ -92,35 +92,48 @@ getProtoFiles := {
...
     import sys.process._
 
-    val releaseUrl = s"https://github.com/bblfsh/libuast/archive/v$libuastVersion.tar.gz"
-    println(s"Downloading libuast from $releaseUrl")
+    val ghUrl = "https://github.com/bblfsh/libuast"
+    val os = if (System.getProperty("os.name").toLowerCase.contains("mac os x")) "darwin" else "linux"
+    val binaryReleaseUrl = s"${ghUrl}/releases/download/v${libuastVersion}/libuast-${os}-amd64.tar.gz"
+    println(s"Downloading libuast binary from ${binaryReleaseUrl}")
+
+    s"curl -sL ${binaryReleaseUrl} -o libuast-bin.tar.gz" #&&
```
creachadair: For future consideration: We can probably simplify this script if we run it with `set -e` and `set -o pipefail` enabled; that would allow us to drop some of the explicit connectives. I'm not clear how easy that is from the shell runner here, though.

.travis.yml
```diff
@@ -1,57 +1,66 @@
...
+        - sudo apt-get install -y binutils
+        - ./sbt assembly
+        - ./sbt "testOnly *Close* *ClientVersion* *SupportedLanguages*"
+        - ./sbt "testOnly org.bblfsh.client.v2.BblfshClientParseTest -- -z \"Decoded UAST after parsing\""
+        - ./sbt "testOnly org.bblfsh.client.v2.BblfshClientParseTest -- -z \"Decoded UAST RootNode\""
+      after_failure: *failure_logs_anchor
 
-after_failure:
- - docker logs bblfsh
+    - name: 'Cross-compile, release & publish to Sonatype'
```
creachadair: I guess much of this is pre-existing, but it seems like we have folded together fetching tools with deployment. Maybe for future consideration, but is it practical for us to fetch and install all our tools as a separate step from actually doing the cross-compilation and assembly of the output? 

For one thing, then we can cache the images more easily (though right now the speed of the build is not the long pole), but maybe more importantly we can isolate installation failures from push failures.

src/main/scala/org/bblfsh/client/v2/BblfshClient.scala
bzz: Yes, exactly - this is part of the old code that was moved forward and going to be used for iterator implementation.

src/main/native/jni_utils.h
bzz: This is an old code, I belive it follow the same that was used in [libuast](https://github.com/bblfsh/libuast/blob/master/src/uast.h).

I'm not settled on this one either, but I think it's totally fine to use almost whatever in a the well-localized JNI wrappers themselves.

Will be happy to refactor it, if needed, and add a better formatter config to all the projects later on, as soon as we have something functional here.  I'm mostly not very worried about it here because of the limited scope and no expectations for this native piece to grow.

TBH Google one from above was just placed randomly, to have at least something, and mostly because of 2-space indentation.

build.sbt
bzz: I agree. Most probably going to replace this this with either shell or a Makefile later on, that will do this work and call into `./sbt`. But would prefer doing that a bit later, after bare-bone functionality of #84 is in place.

.travis.yml
bzz: Totally agree! And we already have an open issue for that work https://github.com/bblfsh/client-scala/issues/87

build.sbt
bzz: Filed https://github.com/bblfsh/client-scala/issues/88 for it not to get lost.

.clang-format
bzz: addressed in f023d7ca8114575f0ee94a928fdc4b81b2544d4d

bzz:
> Just a few questions left.

@creachadair I tried to answer all of those and address the feedback, let me know what you think!
And it's ready for another round.

src/main/native/jni_utils.h
creachadair: Yeah, this doesn't need to be addressed right now, I just wanted to make a note of it while we're paying attention to this code. 🙂

PR #89, opened 2019-06-06
encode() and a stub for the load()

bzz:
Another part of #83

 - `Context.encode()` implementation  + tests
 - a stub for `Node.load()`

Actual load() implementation will be done in subsequent PRs.

### DIFF
--- a/src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc
+++ b/src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc
@@ -15,6 +15,8 @@ jobject asJvmBuffer(uast::Buffer buf) {
   return env->NewDirectByteBuffer(buf.ptr, buf.size);
 }
 
+const char *nativeContext = "nativeContext";
+
 jfieldID getHandleField(JNIEnv *env, jobject obj, const char *name) {
   jclass cls = env->GetObjectClass(obj);
   if (env->ExceptionOccurred() || !cls) {
@@ -145,7 +147,7 @@ JNIEXPORT jobject JNICALL Java_org_bblfsh_client_v2_libuast_Libuast_filter(
 // v2.Context()
 JNIEXPORT jobject JNICALL Java_org_bblfsh_client_v2_Context_root(JNIEnv *env,
                                                                  jobject self) {
-  ContextExt *p = getHandle<ContextExt>(env, self, "nativeContext");
+  ContextExt *p = getHandle<ContextExt>(env, self, nativeContext);
   return p->RootNode();
 }
 
@@ -153,14 +155,14 @@ JNIEXPORT jobject JNICALL Java_org_bblfsh_client_v2_Context_encode(
     JNIEnv *env, jobject self, jobject node) {
   UastFormat fmt = UAST_BINARY;  // TODO(bzz): make it argument & enum
 
-  ContextExt *p = getHandle<ContextExt>(env, self, "nativeContext");
+  ContextExt *p = getHandle<ContextExt>(env, self, nativeContext);
   return p->Encode(node, fmt);
 }
 
 JNIEXPORT void JNICALL Java_org_bblfsh_client_v2_Context_dispose(JNIEnv *env,
                                                                  jobject self) {
-  ContextExt *p = getHandle<ContextExt>(env, self, "nativeContext");
-  setHandle<ContextExt>(env, self, 0, "nativeContext");
+  ContextExt *p = getHandle<ContextExt>(env, self, nativeContext);
+  setHandle<ContextExt>(env, self, 0, nativeContext);
   delete p;
 }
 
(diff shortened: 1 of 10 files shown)


src/main/scala/org/bblfsh/client/v2/JNode.scala
```diff
@@ -0,0 +1,11 @@
+package org.bblfsh.client.v2
+
+sealed abstract class JNode
```
creachadair: Can you please add a comment to say what this is for, even if only briefly?

src/main/scala/org/bblfsh/client/v2/package.scala
```diff
@@ -0,0 +1,5 @@
+package org.bblfsh.client
+
+package object v2 {
```
creachadair: Can you please comment what this is for?

src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc
```diff
@@ -156,6 +149,28 @@ JNIEXPORT jobject JNICALL Java_org_bblfsh_client_v2_Context_root(JNIEnv *env,
   return p->RootNode();
 }
 
+JNIEXPORT jobject JNICALL Java_org_bblfsh_client_v2_Context_encode(
+    JNIEnv *env, jobject self, jobject node) {
+  UastFormat fmt = UAST_BINARY;  // TODO(bzz): make it argument & enum
+
+  ContextExt *p = getHandle<ContextExt>(env, self, "nativeContext");
+  return p->Encode(node, fmt);
+  return nullptr;
```
dennwc:
```suggestion
```

src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc
```diff
@@ -156,6 +149,28 @@ JNIEXPORT jobject JNICALL Java_org_bblfsh_client_v2_Context_root(JNIEnv *env,
...
+}
+
+JNIEXPORT void JNICALL Java_org_bblfsh_client_v2_Context_dispose(JNIEnv *env,
+                                                                 jobject self) {
+  ContextExt *p = getHandle<ContextExt>(env, self, "nativeContext");
+  setHandle<ContextExt>(env, self, 0, "nativeContext");
```
dennwc: Maybe move `"nativeContext"` to a constant?

src/main/scala/org/bblfsh/client/v2/package.scala
bzz: This only serves for defining package-wide type aliases - do you think it is still worth documenting that, even if that is a standard way of doing so in Scala?

Will add a doc for the type alias.

bzz: addressed in 8162713fa88a425510391838a9f51c9bb83c5466

src/main/scala/org/bblfsh/client/v2/JNode.scala
bzz: addressed in 8162713fa88a425510391838a9f51c9bb83c5466

src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc
bzz: addressed in 0d76c98dbdf5ec8db5dedb43dda35a26d5f59c13

bzz:
Thank you for prompt reviews! 
@dennwc all feedback addressed, ready for another round.

src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc
```diff
@@ -15,6 +15,8 @@ jobject asJvmBuffer(uast::Buffer buf) {
   return env->NewDirectByteBuffer(buf.ptr, buf.size);
 }
 
+const char *nativeContext = "nativeContext";
```
creachadair: Minor nit: Either `const char * const nativeContext = "…"` or `const char nativeContext[] = "…"` is preferable here—to ensure the pointer itself is also const (either works). `constexpr char nativeContext[] = "…"` will also work if we assume C++ ≥ 11.

src/main/scala/org/bblfsh/client/v2/package.scala
creachadair: If this is conventional (and would thus not surprise someone actually familiar with Scala, unlike me) then I don't think there's any need to document it specially.

src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc
bzz: Yes, thank you, we do! Was thinking about `constexpr` as well - will replace as part of the next PR.

PR #91, opened 2019-07-01
Node.load() implementation

bzz:
This is a next step from the #83 - loading Go-decoded tree to the JVM memory.

It consists of
  * `Node.load()` implementation + tests
  * better error handling: JVM runtime exceptions instead of `std::runtime_error`
  * restore single CI profile as all existing tests should be passing now
  * small tests refactoring to extract `BblfshClientBaseTest`

---
This change is [Reviewable](https://reviewable.io/reviews/bblfsh/client-scala/91)


src/main/native/jni_utils.cc
```diff
@@ -21,38 +20,130 @@ JNIEnv *getJNIEnv() {
...
-    return NULL;
-  }
+  checkJvmException(std::string("failed to find a class ").append(className));
 
   jmethodID initId = env->GetMethodID(cls, "<init>", initSign);
-  if (env->ExceptionOccurred() || !initId) {
-    return NULL;
-  }
+  checkJvmException(
+      std::string("failed to call <inti> constructor for ").append(className));
```
creachadair:
```suggestion
      std::string("failed to call <init> constructor for ").append(className));
```

src/main/native/jni_utils.cc
```diff
@@ -21,38 +20,130 @@ JNIEnv *getJNIEnv() {
...
+const char *METHOD_JNODE_VALUE_AT = "(I)Lorg/bblfsh/client/v2/JNode;";
+const char *METHOD_JOBJ_ADD =
+    "(Ljava/lang/String;Lorg/bblfsh/client/v2/JNode;)Lscala/collection/"
+    "mutable/Buffer;";
+const char *METHOD_JARR_ADD =
+    "(Lorg/bblfsh/client/v2/JNode;)Lscala/collection/mutable/Buffer;";
+
+const char *METHOD_OBJ_TO_STR = "()Ljava/lang/String;";
+
+void checkJvmException(std::string msg) {
```
creachadair: This function is doing enough interesting work that I think it would benefit from a comment. In particular, it affects control flow in the host program so that probably deserves a call-out IMO.

src/main/native/jni_utils.cc
```diff
@@ -21,38 +20,130 @@ JNIEnv *getJNIEnv() {
...
+  JNIEnv *env = getJNIEnv();
+  if (env->ExceptionCheck()) {
+    // env->ExceptionDescribe(); // prints to stdout, un-comment for debuging
```
creachadair: [optional] Maybe we should leave this in, and put it behind a flag? It can be tricky to repro issues arising from control flow across the FFI boundary, so being able to reload the binary with a debugging flag rather than recompile the whole application may help here (and we can't log in in the "normal" way otherwise at this point in the program). What do you think?

src/main/native/jni_utils.cc
```diff
@@ -21,38 +20,130 @@ JNIEnv *getJNIEnv() {
...
 
   return instance;
 }
 
 jfieldID getField(JNIEnv *env, jobject obj, const char *name) {
   jclass cls = env->GetObjectClass(obj);
-  if (env->ExceptionOccurred() || !cls) {
-    return nullptr;
-  }
+  checkJvmException("failed get an object class");
```
creachadair: Include the class name here too?

src/main/native/jni_utils.cc
```diff
@@ -21,38 +20,130 @@ JNIEnv *getJNIEnv() {
...
+                        .append(".")
+                        .append(method));
+
+  return mId;
+}
+
+jint IntMethod(JNIEnv *env, const char *method, const char *signature,
+               const char *className, const jobject *object) {
+  jmethodID mId = MethodID(env, method, signature, className);
+  checkJvmException(std::string("failed get a method ")
```
creachadair:
```suggestion
  checkJvmException(std::string("failed to get method ")
```

(optional nit)

src/main/native/jni_utils.cc
```diff
@@ -21,38 +20,130 @@ JNIEnv *getJNIEnv() {
...
+                        .append(className)
+                        .append(".")
+                        .append(method));
+
+  jint res = env->CallIntMethod(*object, mId);
+  checkJvmException(std::string("failed call a method ")
```
creachadair:
```suggestion
  checkJvmException(std::string("failed to call method ")
```

(optional nit)

src/main/native/jni_utils.cc
```diff
@@ -21,38 +20,130 @@ JNIEnv *getJNIEnv() {
...
+const char *CLS_CTX = "org/bblfsh/client/v2/Context";
+const char *CLS_OBJ = "java/lang/Object";
+const char *CLS_RE = "java/lang/RuntimeException";
+
+const char *CLS_JNODE = "org/bblfsh/client/v2/JNode";
+const char *CLS_JNULL = "org/bblfsh/client/v2/JNull";
+const char *CLS_JSTR = "org/bblfsh/client/v2/JString";
+const char *CLS_JINT = "org/bblfsh/client/v2/JInt";
+const char *CLS_JFLT = "org/bblfsh/client/v2/JFloat";
+const char *CLS_JBOL = "org/bblfsh/client/v2/JBool";
```
creachadair:
```suggestion
const char *CLS_JBOOL = "org/bblfsh/client/v2/JBool";
```

(I think?)

src/main/native/jni_utils.h
```diff
@@ -2,12 +2,40 @@
 #define _Included_org_bblfsh_client_libuast_Libuast_jni_utils
 
 #include <jni.h>
+#include <string>
 
 extern const char *CLS_NODE;
 extern const char *CLS_CTX;
+extern const char *CLS_OBJ;
```
creachadair: Some of this is pre-existing, but consider documenting these names here in the header (this is what will be included by anywhere it's used, so this is presumably what the reader is going to look at).

A the least, please consider leaving a block comment nearby these definitions, to describe what they're used for. Some of the names are kind of obvious, but for example it isn't obvious that these are Babelfish types rather than Java ones, for example.

src/main/scala/org/bblfsh/client/v2/JNode.scala
```diff
@@ -1,15 +1,60 @@
...
-case class JObject(obj: List[JField]) extends JNode
-case class JArray(arr: List[JNode]) extends JNode
+sealed abstract class JNode {
+  def children: Seq[JNode] = this match {
+    case JObject(l) => l map (_._2)
+    case JArray(l) => l
+    case _ => Seq()
+  }
+
+  def size: Int = this match {
```
creachadair: I don't know what's idiomatic for Scala, but it seems slightly odd to do this as a dispatch on type in the base class rather than as an override in the subclasses with a default here. If the subclasses don't add any behaviour, do they need to be related at all?

src/main/scala/org/bblfsh/client/v2/JNode.scala
```diff
@@ -1,15 +1,60 @@
...
+    case JObject(l) => l.size
+    case JArray(l) => l.size
+    case JString(l) => l.length
+    case JNothing => 0
+    case _ => 1
```
creachadair: What does 1 signify here? (And why 1 instead of, say, error?)

src/main/scala/org/bblfsh/client/v2/JNode.scala
bzz: In general, yes, that is true, but in our particular case - this is just a convenience, in order to avoid having this exact logic on JNI side.

src/main/scala/org/bblfsh/client/v2/Context.scala
```diff
@@ -1,13 +1,15 @@
...
+
 case class Context(nativeContext: Long) {
+    def this() = this(Context.create())
+
     @native def root(): Node
     @native def encode(n: Node): ByteBuffer
     @native def dispose()
 
-    @native def filter()
+    @native def filter() // TODO(bzz)
```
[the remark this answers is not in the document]
bzz: No, sorry for confusion - this is literally a mark for the next step, that I'm already working on for the next PR.
If that is alright, I would prefer to remove it in there.

src/main/native/jni_utils.cc
bzz: I see what you mean, but think that I would prefer to simply remove this, if anybody feel strongly agains having this comment here.

Having a flag + reloading the native library from the JVM runtime sounds too error-prone and I do not think it will worth it. Especially, given the fact that one has to rebuild the native part with the clang debug flag anyway, in order to preserve the symbols.

src/main/scala/org/bblfsh/client/v2/JNode.scala
bzz: It signifies a single element rather then a container one, and you are right, it should be 0 instead (same as in other clients).

src/main/native/jni_utils.cc
bzz: We do not have a name of the class at this point.

bzz:
@creachadair All feedback addressed in 1e7858f, ready for another round!

bzz:
I'm a bit lost with reviewable as we've been mixing it with GH reviews here, but @dennwc  - all feedback was addressed, so please take a look.

This and #93 blocks further work as all changes go to the same place, so would prefer merge it soon.

PR #105, opened 2019-07-31
Fix encoding for JUint 

### LINKED ISSUE #104: Encoding UAST back does not produce the same bytes
A new test case when a decode UAST serialized though a new, different from original one, context does not result in the same bytes:

```scala
    // given
    val resp = client.parse(fileName, fileContent)
    val node = resp.get

    // when
    var encodedBytes = node.toByteBuffer

    // then
    encodedBytes.capacity should be(resp.uast.asReadOnlyByteBuffer.capacity)
    encodedBytes shouldEqual resp.uast.asReadOnlyByteBuffer
```

In particular, this seems to be the case for JUint nodes.

bzz:
Fixes #104 

 - rename var in native part to avoid type confusion
 - simplify test fixture for python
 - add new tests for decode->new context->encode case

---
This change is [Reviewable](https://reviewable.io/reviews/bblfsh/scala-client/105)


bzz:
Thank you for prompt reviews! Going to merge this and 🔪 a new RC3 release asap

Issue #113, opened 2019-08-14
Node::ValueAt may be leaking memory


ncordon:
After a glimpse to the methods in `Node` (C++ side), I have the (untested of course) hypothesis that [`ValueAt`](https://github.com/bblfsh/scala-client/blob/ee89f22091c1bfea42037063856e40be33c040eb/src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc#L335) may be leaking memory.

* `ValueAt` creates new global ref for and object `obj` and calls `lookupOrCreate(obj)`
* `lookupOrCreate` tries to find a `Node` for `obj` (if it previously existed, we created a new reference to it, and we are only deallocating one). If no `Node` existed, calls the constructor of `Node`, which creates a new reference to `obj`, but we do not release the original one at any point D:

Proposed solution is end `ValueAt` method with:

```
return lookupOrCreate(val);
```

instead of

```
return lookupOrCreate(env->NewGlobalRef(val));
```

bzz:
This is a good observation, but merely implementing the suggestion will result in test failures.

I have submitted a patch for what seems to be one of the causes of this at #116 (cann't use refs as keys for the map really) but it is still a partial solution - if global ref is removed there, it will fail the `PreOrder` managed iterator test :/ 

I suggest also adding to the scope of this issue identifying the reason of why does `Interface::toJ` has to do `getJNIEnv()->NewGlobalRef()`?  

It does not seem to need one for the same reasons, doesn't it? But removing it causes `BblfshClientLoadTest` to fail with `java.lang.AbstractMethodError: org.bblfsh.client.v2.JNode.size()I` on `Context.encode`.

I hope that new materials from #115 on JNI debugging at [CONTRIBUTING.m](https://github.com/bblfsh/scala-client/blob/79614f9932e65e1aceec340e87ff7b7ec3381b4e/CONTRIBUTING.md#developer-documentation) will help to find an answer.

IMO it makes sense to start with this one and when fixed 🔪  a release, so added it to [v2.0.1 milestone](https://github.com/bblfsh/scala-client/milestone/4).

ncordon:
Closed by #116

PR #115, opened 2019-08-15
Docs and leaks

bzz:
Addresses the last part of the #86 needed for the release:

TODOs:
 - [x] small JNI utilities API improvement
 - [x] README cleanup
 - [x] ~address part of the #113 by changing native map cache implementation~
           moved to separate #116 
 - [x] avoid leaking memory on managed iterators by not passing GlobalRefs
 - [x] add CONTRIBUTING.md \w additional developer docs

---
This change is [Reviewable](https://reviewable.io/reviews/bblfsh/scala-client/115)

PR #116, opened 2019-08-16
Avoid using JNI refs as map keys

bzz:
This is a proposition for addressing the #113.

Original impl used a reference as a key in a native map, but in JNI object references are neither constant nor unique, and thus were violating the contract on native map `hashCode()`-equivalent for a key.

This should allow to get rid of global reference in the interface lookup, but there more (clearly identified) cases where we still rely on side-effects of having a global reference and thus leak some memory.

This seems to works on macOS but fails on linux, but needs further work:
  - this cahnge fails on linux, but works on macOS 
  - when `NewGloablRef` in interface lookup is removed, it a pre-order iterator tests

This branch is intentionally pushed to upstream repo, so @ncordon please feel free to take over and help.

---
This change is [Reviewable](https://reviewable.io/reviews/bblfsh/scala-client/116)

### DIFF
--- a/src/main/native/jni_utils.cc
+++ b/src/main/native/jni_utils.cc
@@ -61,12 +61,12 @@ const char METHOD_RE_INIT_CAUSE[] =
 const char METHOD_ITER_INIT[] = "(Lorg/bblfsh/client/v2/NodeExt;IJLorg/bblfsh/client/v2/ContextExt;)V";
 const char METHOD_JITER_INIT[] = "(Lorg/bblfsh/client/v2/JNode;IJLorg/bblfsh/client/v2/Context;)V";
 
+const char METHOD_NODE_INIT[] = "(Lorg/bblfsh/client/v2/ContextExt;J)V";
+
 // Field signatures
 const char FIELD_ITER_NODE[] = "Ljava/lang/Object;";
-const char FIELD_ITER_CTX[] = "Lorg/bblfsh/client/v2/Context;";
-const char FIELD_ITER_EXT_NODE[] = "Ljava/lang/Object;";
-const char FIELD_ITER_EXT_CTX[] = "Lorg/bblfsh/client/v2/ContextExt;";
-const char FIELD_NODE_EXT_CTX[] = "Lorg/bblfsh/client/v2/ContextExt;";
+const char FIELD_CTX[] = "Lorg/bblfsh/client/v2/Context;";
+const char FIELD_CTX_EXT[] = "Lorg/bblfsh/client/v2/ContextExt;";
 
 // TODO(#114): cache classes&methods in JNI_OnLoad should speed this up
 void checkJvmException(std::string msg) {
--- a/src/main/native/jni_utils.cc
+++ b/src/main/native/jni_utils.cc
@@ -26,7 +26,8 @@ JNIEnv *getJNIEnv() {
 
 // Class fully qualified names
 const char CLS_NODE[] = "org/bblfsh/client/v2/NodeExt";
-const char CLS_CTX[] = "org/bblfsh/client/v2/ContextExt";
+const char CLS_CTX_EXT[] = "org/bblfsh/client/v2/ContextExt";
+const char CLS_CTX[] = "org/bblfsh/client/v2/Context";
 const char CLS_TO[] = "org/bblfsh/client/v2/TreeOrder";
 const char CLS_OBJ[] = "java/lang/Object";
 const char CLS_RE[] = "java/lang/RuntimeException";
@@ -57,11 +58,15 @@ const char METHOD_RE_INIT[] = "(Ljava/lang/String;)V";
 const char METHOD_RE_INIT_CAUSE[] =
     "(Ljava/lang/String;Ljava/lang/Throwable;)V";
 
-const char METHOD_ITER_INIT[] = "(Lorg/bblfsh/client/v2/NodeExt;IJJ)V";
-const char METHOD_JITER_INIT[] = "(Lorg/bblfsh/client/v2/JNode;IJJ)V";
+const char METHOD_ITER_INIT[] = "(Lorg/bblfsh/client/v2/NodeExt;IJLorg/bblfsh/client/v2/ContextExt;)V";
+const char METHOD_JITER_INIT[] = "(Lorg/bblfsh/client/v2/JNode;IJLorg/bblfsh/client/v2/Context;)V";
 
 // Field signatures
 const char FIELD_ITER_NODE[] = "Ljava/lang/Object;";
+const char FIELD_ITER_CTX[] = "Lorg/bblfsh/client/v2/Context;";
+const char FIELD_ITER_EXT_NODE[] = "Ljava/lang/Object;";
+const char FIELD_ITER_EXT_CTX[] = "Lorg/bblfsh/client/v2/ContextExt;";
+const char FIELD_NODE_EXT_CTX[] = "Lorg/bblfsh/client/v2/ContextExt;";
 
 // TODO(#114): cache classes&methods in JNI_OnLoad should speed this up
 void checkJvmException(std::string msg) {
--- a/src/main/native/jni_utils.h
+++ b/src/main/native/jni_utils.h
@@ -36,14 +36,12 @@ extern const char METHOD_RE_INIT[];
 extern const char METHOD_RE_INIT_CAUSE[];
 extern const char METHOD_ITER_INIT[];
 extern const char METHOD_JITER_INIT[];
+extern const char METHOD_NODE_INIT[];
 
 // Field signatures
 extern const char FIELD_ITER_NODE[];
-extern const char FIELD_ITER_EXT_NODE[];
-extern const char FIELD_NODE_EXT_CTX[];
-extern const char FIELD_ITER_CTX[];
-extern const char FIELD_ITER_EXT_CTX[];
-
+extern const char FIELD_CTX[];
+extern const char FIELD_CTX_EXT[];
 // Checks through JNI, if there is a pending excption on the JVM side.
 //
 // Throws new RuntimeException to the JVM in case there is,
--- a/src/main/native/jni_utils.h
+++ b/src/main/native/jni_utils.h
@@ -6,6 +6,8 @@
 
 // Fully qualified Java class names
 extern const char CLS_NODE[];
+extern const char CLS_CTX_EXT[];
+extern const char CLS_CTX[];
 extern const char CLS_CTX[];
 extern const char CLS_OBJ[];
 extern const char CLS_RE[];
@@ -37,6 +39,10 @@ extern const char METHOD_JITER_INIT[];
 
 // Field signatures
 extern const char FIELD_ITER_NODE[];
+extern const char FIELD_ITER_EXT_NODE[];
+extern const char FIELD_NODE_EXT_CTX[];
+extern const char FIELD_ITER_CTX[];
+extern const char FIELD_ITER_EXT_CTX[];
 
 // Checks through JNI, if there is a pending excption on the JVM side.
 //
--- a/src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc
+++ b/src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc
@@ -144,7 +144,7 @@ class ContextExt {
     delete (ctx);
 
     if (jCtxExt)
-      getJNIEnv()->DeleteGlobalRef(jCtxExt);
+      getJNIEnv()->DeleteWeakGlobalRef(jCtxExt);
   }
 
   // lookup searches for a specific node handle.
@@ -159,7 +159,7 @@ class ContextExt {
   // We need this because a NodeExt from Scala side includes
   // a Scala ContextExt and a handle to the native C node
   void setJavaContext(jobject ctx) {
-    jCtxExt = getJNIEnv()->NewGlobalRef(ctx);
+    jCtxExt = getJNIEnv()->NewWeakGlobalRef(ctx);
   }
     
   // Iterate returns iterator over an external UAST tree.
--- a/src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc
+++ b/src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc
@@ -365,7 +365,7 @@ class Node : public uast::Node<Node *> {
     jobject val =
         ObjectMethod(env, "valueAt", METHOD_JNODE_VALUE_AT, CLS_JNODE, obj, i);
     // TODO(#113) investigate, it looks like a potential memory leak
-    return lookupOrCreate(env->NewGlobalRef(val));  // new ref
+    return lookupOrCreate(val);  // borrows the reference
   }
 
   void SetValue(size_t i, Node *val) {
@@ -430,12 +430,13 @@ class Interface : public uast::NodeCreator<Node *> {
   Node *lookupOrCreate(jobject obj) {
     if (!obj) return nullptr;
 
-    Node *node = obj2node[obj];
-    if (node) return node;
-
-    node = new Node(this, obj);
-    obj2node[node->obj] = node;
-    return node;
+    if (obj2node.count(obj) > 0) {
+      return obj2node[obj];
+    } else {
+      Node* node = new Node(this, obj);
+      obj2node[node->obj] = node;
+      return node;
+    }
   }
 
   // create makes a new object with a specified kind.
--- a/src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc
+++ b/src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc
@@ -51,6 +51,7 @@ void setHandle(JNIEnv *env, jobject obj, T *t, const char *name,
 
   jlong handle = reinterpret_cast<jlong>(t);
   env->SetLongField(obj, fId, handle);
+  checkJvmException("failed to set handle for" + std::string(name));
 }
 
 template <typename T>
@@ -61,6 +62,7 @@ void setHandle(JNIEnv *env, jobject obj, T *t, const char *name) {
 void setObjectField(JNIEnv *env, jobject obj, jobject field, const char *name, const char *sig) {
   jfieldID fId = FieldID(env, obj, name, sig);
   env->SetObjectField(obj, fId, field);
+  checkJvmException("failed to set object field for" + std::string(name));
 }
 
 jobject asJvmBuffer(uast::Buffer buf) {
@@ -103,7 +105,7 @@ class ContextExt {
     if (node == 0) return nullptr;
 
     JNIEnv *env = getJNIEnv();
-    jobject jObj = NewJavaObject(env, CLS_NODE, "(Lorg/bblfsh/client/v2/ContextExt;J)V", jCtxExt, node);
+    jobject jObj = NewJavaObject(env, CLS_NODE, METHOD_NODE_INIT, jCtxExt, node);
     return jObj;
   }
 
@@ -158,7 +160,7 @@ class ContextExt {
   // Attaches a Scala ContextExt object to the C ContextExt
   // We need this because a NodeExt from Scala side includes
   // a Scala ContextExt and a handle to the native C node
-  void setJavaContext(jobject ctx) {
+  void setManagedContext(jobject ctx) {
     jCtxExt = getJNIEnv()->NewWeakGlobalRef(ctx);
   }
 
@@ -606,9 +608,8 @@ class Context {
 
   jobject LoadFrom(jobject src) {  // NodeExt
     JNIEnv *env = getJNIEnv();
-    // NodeExt contains a ctx: ContextExt and ContextExt the
-    // handle for the native context, called nativeContext
-    jobject jCtxExt = ObjectField(env, src, "ctx", FIELD_NODE_EXT_CTX);
+    // NodeExt contains a ctx: ContextExt (JVM ref) and a nativeContext: ContextExt (handle)
+    jobject jCtxExt = ObjectField(env, src, "ctx", FIELD_CTX_EXT);
     ContextExt *nodeExtCtx = getHandle<ContextExt>(env, jCtxExt, nativeContext);
 
     checkJvmException("failed to get NodeExt.ctx");
@@ -650,8 +651,8 @@ JNIEXPORT jobject JNICALL Java_org_bblfsh_client_v2_libuast_Libuast_decode(
 
   jobject jCtxExt = NewJavaObject(env, CLS_CTX_EXT, "(J)V", p);
 
-  // Associates the JVM context ext to the native ContextExt
-  p->setJavaContext(jCtxExt);
+  // Saves weak reference to JVM ContextExt in the native ContextExt
+  p->setManagedContext(jCtxExt);
 
   if (env->ExceptionCheck() || !jCtxExt) {
     jCtxExt = nullptr;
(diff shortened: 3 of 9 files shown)


bzz:
CI passes now, before rebasing on #115

bzz:
Rebased on latest master (after was #115 merged).

ncordon:
I have get rid (I think) of the memory leaks from #113. I think maybe it would make sense to remove all the comments of new reference, steal reference, because there is no such thing here as it is Python (all the references have to be cleaned up by the JNI, either explicitly if they are global refs, or implicitly as in local refs, which are cleaned automatically by JNI). WDYT?

ncordon:
I am going to block my own PR because I found a big memory leak this afternoon (simply parsing a file in a while similar as to what I did with the Python client). The fix for the memory leak is not useful if I do not do some further changes (I have to think how). The memory leak was caused because the `ContextExt` dispose method in JNI had a condition flipped, and was never deleting the tree. I am going to try and explain the problem:

Consider the following code:

```
import scala.io.Source
import org.bblfsh.client.v2.{BblfshClient, NodeExt}, BblfshClient._
import gopkg.in.bblfsh.sdk.v2.protocol.driver.Mode

val client = BblfshClient("localhost", 9432)

val fileName = "src/test/resources/python_file.py"

val resp = client.parse(fileName, fileContent)
val uast = resp.uast.decode()
val rootNode: NodeExt = uast.root()
val root = rootNode.load()
```

- `uast` is a `ContextExt` which holds an integer which points to the native context (in JNI).
- When I do `val rootNode = uast.root()`, I generate a `NodeExt`, which again holds a pointer to the native context and to the handle of the node.
- When I do a `rootNode.load()`, I should not have disposed of the `ContextExt`, since otherwise I am going to try an access freed memory.

Problem is that `NodeExt` should not hold a pointer to the native context, but to `ContextExt`. If not, the JVM Garbage Collector could get in between those calls and free `uast` (because there is not anything else pointing to it). This can be checked with:

```
val uast = resp.uast.decode()
val rootNode: NodeExt = uast.root()
uast.dispose()
val root = rootNode.load()
```

and we end up with a SIGSEV.

Before doing my commit and flipping the condition we had SIGSEVs in `Context`, because it is similar to `ContextExt` but this one was actually deleting the underlying native data. That's why I started digging into this, because I was getting cores from time to time running the tests (hardly ever, because JVM GC has to be triggered in a very precise point).

ncordon:
___
*[src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc, line 368 at r2](https://reviewable.io/reviews/bblfsh/scala-client/116#-Lmp5QRgDp7ie5NIgIWJ:-LmpPsC38F8fnFslwVBh:be9aeqj) ([raw file](https://github.com/bblfsh/scala-client/blob/bf90815b7bd8c63e797de070f1ad3739585dedb5/src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc#L368)):*

Depending on the resolution of @ncordon's question: If we _are_ borrowing a pointer here I think that should be part of the function's doc comment, rather than a comment on the `return` statement, since that's something the caller will need to be aware of.
</blockquote></details>

I am going to restructure docs then

ncordon:
___
*[src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc, line 435 at r2](https://reviewable.io/reviews/bblfsh/scala-client/116#-Lmp5k2-7jHpcEQqpUuC:-LmpQ4PV60iSkm05Fpbf:bsiqen0) ([raw file](https://github.com/bblfsh/scala-client/blob/bf90815b7bd8c63e797de070f1ad3739585dedb5/src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc#L435)):*

I suggest treating the cache hit as an early return case:

```c
if (obj2node.count(obj) > 0) {
   return obj2node[obj];
}
Node *node  = new Node(this, obj);
// etc.
```

Relatedly: Please be consistent about pointer/name grouping. It looks like most of this file follows the traditional "separate" convention (`A *a` vs. `A* a`). 

(That's the style I personally prefer too, because of how it binds in the grammar, but I don't actually care as long as we are reasonably consistent)

</blockquote></details>

I will fix this, and yes, I'll try to be consistent with the current style :)

ncordon:
___
*[src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc, line 100 at r4](https://reviewable.io/reviews/bblfsh/scala-client/116#-LnFniPOC9Gn5B4kJet1:-LnFniPPCmjP0k75urLk:bfz3awh) ([raw file](https://github.com/bblfsh/scala-client/blob/f1048d102ea0202ee337d05f19df284ed0661454/src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc#L100)):*
> ```cpp
>  private:
>   uast::Context<NodeHandle> *ctx;
>   jobject jCtxExt;
> ```

Rationale behind this is that I have to tie the `ContextExt` in the C++ side to the `jCtx: ContextExt` Scala object. That way every time that we ask for a node and we are returned a `NodeExt`, it can box `jCtx` so GC thinks: okay, jCtx is being used by a `NodeExt` so I cannot deallocate it. Former implementation only included the handle (a pointer) in the Scala`NodeExt`. That did not prevent GC from collecting `jCtx` before we ended up working with our node (for example doing iterators). Same criteria applies to iterators, which also box a `Context / ContextExt` now.

ncordon:
___
*[src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc, line 163 at r4](https://reviewable.io/reviews/bblfsh/scala-client/116#-LnFokTH0Fia_yAp6GXY:-LnFokTH0Fia_yAp6GXZ:b-9zx308) ([raw file](https://github.com/bblfsh/scala-client/blob/f1048d102ea0202ee337d05f19df284ed0661454/src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc#L163)):*
> ```cpp
>   void setJavaContext(jobject ctx) {
>     jCtxExt = getJNIEnv()->NewWeakGlobalRef(ctx);
>   }
> ```

`jCtxExt` being a `WeakReference` makes possible for GC to collect `jCtxExt` from outside the JNI if the only thing that is using it is the native `ContextExt`. If not, we would end up with an auto-reference to deallocate: Scala `ContextExt` has a handle to a native `ContextExt` which also contains `JCtxExt`, with the result of never deallocating anything

ncordon:
___
*[src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc, line 613 at r4](https://reviewable.io/reviews/bblfsh/scala-client/116#-LnFrVoYBF2CaEEqsr0s:-LnFrVoYBF2CaEEqsr0t:b-p929z7) ([raw file](https://github.com/bblfsh/scala-client/blob/5b0972db1352b88634e11a82f867ab35c73fdbcd/src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc#L613)):*
> ```cpp
>   if (env->ExceptionCheck() || !jCtxExt) {
>     jCtxExt = nullptr;
>     delete (ctx);
> ```

`p` also deletes `ctx`

ncordon:
___
*[src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc, line 819 at r4](https://reviewable.io/reviews/bblfsh/scala-client/116#-LnFrdWF-UMSMekH6ZGi:-LnFrdWF-UMSMekH6ZGj:b-o8zpno) ([raw file](https://github.com/bblfsh/scala-client/blob/f1048d102ea0202ee337d05f19df284ed0661454/src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc#L819)):*
> ```cpp
>   // new UastIter()
>   jobject iter =
>       NewJavaObject(env, CLS_JITER, METHOD_JITER_INIT, 0, 0, it, self);
> ```

Note that now the iterators do not contain only the opaque handle to the context, but a reference to the Java objet which contains the handle, to avoid deallocations of that context.

ncordon:
___
*[src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc, line 819 at r4](https://reviewable.io/reviews/bblfsh/scala-client/116#-LnFrdWF-UMSMekH6ZGi:-LnFrqcq4zq9uypWuZ51:b-47ajlo) ([raw file](https://github.com/bblfsh/scala-client/blob/f1048d102ea0202ee337d05f19df284ed0661454/src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc#L819)):*

Note that now the iterators do not contain only the opaque handle to the context, but a reference to the Java objet which contains the handle, to avoid deallocations of that context.
</blockquote></details>

*object

ncordon:
___
*[src/main/scala/org/bblfsh/client/v2/ContextExt.scala, line 36 at r4](https://reviewable.io/reviews/bblfsh/scala-client/116#-LnFs-xU6QOP7_A7Imzu:-LnFs-xU6QOP7_A7Imzv:b6yadq8) ([raw file](https://github.com/bblfsh/scala-client/blob/f1048d102ea0202ee337d05f19df284ed0661454/src/main/scala/org/bblfsh/client/v2/ContextExt.scala#L36)):*
> ```Scala
>     override def finalize(): Unit = {
>       this.dispose()
>     }
> ```

Note the finalizer for this class was missing

ncordon:
___
*[src/test/scala/org/bblfsh/client/v2/libuast/IteratorNativeTest.scala, line 48 at r4](https://reviewable.io/reviews/bblfsh/scala-client/116#-LnFsAme5bGEMwMEd42J:-LnFsAmf2zT9GUelHDI5:b4fdj62) ([raw file](https://github.com/bblfsh/scala-client/blob/f1048d102ea0202ee337d05f19df284ed0661454/src/test/scala/org/bblfsh/client/v2/libuast/IteratorNativeTest.scala#L48)):*
> ```Scala
> 
>     iter.hasNext() should be(false)
>     iter.ctx should be(null)
> ```

This is ugly Scala, I know, since null references should be avoided. But before this was a handle and null value was 0. Here either we clean the values with `ContextExt(0)` or assign null. I am inclined for the second option

ncordon:
This is ready for review now. I have sanitized as much as possible SIGSEVs, memory leaks and so on

bzz:
Can not add myself to reviewer box in GH UI

ncordon:
Probably because you opened the PR? Do not worry, I will not merge this until you approve changes

ncordon:
___
*[src/main/native/jni_utils.cc, line 69 at r5](https://reviewable.io/reviews/bblfsh/scala-client/116#-LnHNXHSEbOn4Msdde2D:-LnHhulG-d12i8NNrWPA:b-58nxt3) ([raw file](https://github.com/bblfsh/scala-client/blob/72b1d5ad9e3f22b5b212d4f4c95e5d065604614d/src/main/native/jni_utils.cc#L69)):*
<details><summary><i>Previously, bzz (Alexander) wrote…</i></summary><blockquote>

may be one constant could be enough here, since the signatures are the same?
</blockquote></details>

I am going to name fields but what they return and not the class they are applied to

ncordon:
___
*[src/main/native/jni_utils.cc, line 69 at r5](https://reviewable.io/reviews/bblfsh/scala-client/116#-LnHNXHSEbOn4Msdde2D:-LnHi4IM7-h7id1Tp_6Q:be1nrf4) ([raw file](https://github.com/bblfsh/scala-client/blob/72b1d5ad9e3f22b5b212d4f4c95e5d065604614d/src/main/native/jni_utils.cc#L69)):*

I am going to name fields but what they return and not the class they are applied to
</blockquote></details>

*by what they return

ncordon:
___
*[src/main/native/jni_utils.cc, line 67 at r5](https://reviewable.io/reviews/bblfsh/scala-client/116#-LnGNudP8qplqQM8xjJu:-LnHhGTA6M6RIs1NYN4u:b8iao0l) ([raw file](https://github.com/bblfsh/scala-client/blob/72b1d5ad9e3f22b5b212d4f4c95e5d065604614d/src/main/native/jni_utils.cc#L67)):*

Really Object rather than Node or NodeExt? (That's fine if intended, but might benefit from a comment)
</blockquote></details>

Yes, that was intentional. It was @bzz who first noticed it, but Scala classes for iterators extend an abstract class that is "templated" in a type: `abstract class UastAbstractIter[T >: Null](var node: T, var treeOrder: Int, var iter: Long)` and `node` must be accessed as an arbitrary object. Probably because Scala is compiling this in Java as a type that both `NodeExt` and `Node` extend

ncordon:
___
*[src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc, line 63 at r5](https://reviewable.io/reviews/bblfsh/scala-client/116#-LnHNwv8BjA1YcpEiu2m:-LnHknSQ7_QEd8YUqpIO:b-mydzw3) ([raw file](https://github.com/bblfsh/scala-client/blob/72b1d5ad9e3f22b5b212d4f4c95e5d065604614d/src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc#L63)):*
<details><summary><i>Previously, bzz (Alexander) wrote…</i></summary><blockquote>

We should probably also either check for JNI error/exception here, or at every client of this method with `checkJvmException`.
</blockquote></details>

Done! Also added the check to `setHandle` function

ncordon:
___
*[src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc, line 106 at r5](https://reviewable.io/reviews/bblfsh/scala-client/116#-LnHOcWS-Nabmt8aWixZ:-LnHlNVbE_jadVwj4diu:bs7wub4) ([raw file](https://github.com/bblfsh/scala-client/blob/72b1d5ad9e3f22b5b212d4f4c95e5d065604614d/src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc#L106)):*
<details><summary><i>Previously, bzz (Alexander) wrote…</i></summary><blockquote>

probably nit pick, but as soon as signature gets longer and becomes very specific - it might be worth extracting a named constant there
</blockquote></details>

Done!

ncordon:
___
*[src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc, line 163 at r4](https://reviewable.io/reviews/bblfsh/scala-client/116#-LnFokTH0Fia_yAp6GXY:-LnHmHx63wedJhVR8Cdq:bdj6lqs) ([raw file](https://github.com/bblfsh/scala-client/blob/f1048d102ea0202ee337d05f19df284ed0661454/src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc#L163)):*
<details><summary><i>Previously, bzz (Alexander) wrote…</i></summary><blockquote>

As this is a Scala project - a minor suggestion would be to avoiding using `Java` until it really means something languages-specific to void Scala/Java compatibility confusion.

One terminology distinction I found useful: instead of constantly confusing C/C++ vs Scala/Java, that is a _native_ part VS _managed_ part that works for multiple client implementations. Or just JVM to refer a particular managed environment.

One side-effect that may be worth documenting is - `jCtxExt` may now have a valid reference (not `null`) that points to the JVM null object. So every native client consuming this (if we have one) might need to do `env->IsSameObject(jCtxExt, null)`.

It's even a bit more involved - every native client would need to get actually local ref from a weak one, even if he did made the above check - the weak ref might be GCed from another thread after the check and before the usage and thus will refer an null JVM object anyway.

</blockquote></details>

I have changed the name of the method to `setManagedContext`. Thanks for the suggestion.

I have a question regarding the second part because I am not sure it can happen that `jCtxExt` is not null and the object it points to is. I mean, the Scala `jCtx: Context` is associated to a native `ctx: Context` which contains the weak reference `jCtxExt = jCtx`. If `jCtx` points null, it means we have called dispose on it, and therefore we have deleted `ctx` and `jCtxExt`.  The only way we can access `jCtxExt` and point to null is if we manipulate it (copy it for example and then call dispose in the managed side) from the C++ side, isn't? If we return it to the client wrapped in a `NodeExt` for example it stops being a weak reference.

Probably I misunderstood what you said, so feel free to correct me please :)

ncordon:
___
*[src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc, line 435 at r2](https://reviewable.io/reviews/bblfsh/scala-client/116#-Lmp5k2-7jHpcEQqpUuC:-LnHpEHHDVFsF8Z1oBNq:b-wdbe2x) ([raw file](https://github.com/bblfsh/scala-client/blob/bf90815b7bd8c63e797de070f1ad3739585dedb5/src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc#L435)):*

I will fix this, and yes, I'll try to be consistent with the current style :)
</blockquote></details>

This was changed already :)

ncordon:
___
*[src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc, line 368 at r2](https://reviewable.io/reviews/bblfsh/scala-client/116#-Lmp5QRgDp7ie5NIgIWJ:-LnHpH7K9xUbPAGfnLm4:b-896fja) ([raw file](https://github.com/bblfsh/scala-client/blob/bf90815b7bd8c63e797de070f1ad3739585dedb5/src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc#L368)):*

I am going to restructure docs then
</blockquote></details>

Done!

ncordon:
___
*[src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc, line 611 at r5](https://reviewable.io/reviews/bblfsh/scala-client/116#-LnHV9Qi8xnBLYdXqMtZ:-LnHpQ-50tW6Rh-XtlMe:bxa5kst) ([raw file](https://github.com/bblfsh/scala-client/blob/72b1d5ad9e3f22b5b212d4f4c95e5d065604614d/src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc#L611)):*
> // NodeExt contains a ctx: ContextExt (JVM ref) and a nativeContext: ContextExt (handle)
Yeah, much better

ncordon:
___
*[src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc, line 611 at r5](https://reviewable.io/reviews/bblfsh/scala-client/116#-LnHV9Qi8xnBLYdXqMtZ:-LnHpSpp0UxovZl1IeCP:b-y6yk4k) ([raw file](https://github.com/bblfsh/scala-client/blob/72b1d5ad9e3f22b5b212d4f4c95e5d065604614d/src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc#L611)):*

> // NodeExt contains a ctx: ContextExt (JVM ref) and a nativeContext: ContextExt (handle)
Yeah, much better
</blockquote></details>

Changed now!

ncordon:
___
*[src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc, line 653 at r5](https://reviewable.io/reviews/bblfsh/scala-client/116#-LnHTGEhBCv3IfUiCI1K:-LnHppjs-e2Vh2T2Tx7X:b-4ipyfq) ([raw file](https://github.com/bblfsh/scala-client/blob/72b1d5ad9e3f22b5b212d4f4c95e5d065604614d/src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc#L653)):*
> // Saves weak reference to JVM ContextExt in the native ContextExt

Done!

ncordon:
___
*[src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc, line 613 at r4](https://reviewable.io/reviews/bblfsh/scala-client/116#-LnFrVoYBF2CaEEqsr0s:-LnHps-GDAH_3nL-HOnG:b-auyswk) ([raw file](https://github.com/bblfsh/scala-client/blob/5b0972db1352b88634e11a82f867ab35c73fdbcd/src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc#L613)):*

Maybe put that directly into the comment? (e.g., `// Since p owns the underlying context, deleting p deletes that too.`)
</blockquote></details>

This is strange, because the comment is there, but Reviewable is eating that line:

```
  if (env->ExceptionCheck() || !jCtxExt) {
    jCtxExt = nullptr;
    // This also deletes the underlying ctx
    delete (p);
    checkJvmException("failed to instantiate ContextExt class");
  }
```

ncordon:
___
*[src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc, line 740 at r5](https://reviewable.io/reviews/bblfsh/scala-client/116#-LnHVo4SB5JMXmd9-F9K:-LnHq9FtEdVq8P1AZe1w:bogoerq) ([raw file](https://github.com/bblfsh/scala-client/blob/72b1d5ad9e3f22b5b212d4f4c95e5d065604614d/src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc#L740)):*
<details><summary><i>Previously, bzz (Alexander) wrote…</i></summary><blockquote>

Is there a reason no to follow the convention of not having empty line between getting the value and an error check here?
</blockquote></details>

None, only myself not being consistent :laughing:

ncordon:
___
*[src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc, line 847 at r5](https://reviewable.io/reviews/bblfsh/scala-client/116#-LnGRF3V9QihehqGb2h_:-LnHqTtnED1VLi4SEPX0:bnq6tfb) ([raw file](https://github.com/bblfsh/scala-client/blob/72b1d5ad9e3f22b5b212d4f4c95e5d065604614d/src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc#L847)):*

Is there a race condition here where JVM might access the pointer that has just been deleted before we finish updating the field? (I don't know that there is—but if so this could be use-after-free; I think nulling the handle first would avert that problem)
</blockquote></details>

Yeah good advice. I do not know if there is, but as a matter of fact we cannot put `dispose` as private methods, so it is still worth nulling the field.

ncordon:
___
*[src/main/native/project/build.properties, line 1 at r5](https://reviewable.io/reviews/bblfsh/scala-client/116#-LnHWhnmEyF2cCq6gohz:-LnHqxL9DEsVA8JvrJGL:b-jjenie) ([raw file](https://github.com/bblfsh/scala-client/blob/72b1d5ad9e3f22b5b212d4f4c95e5d065604614d/src/main/native/project/build.properties#L1)):*
<details><summary><i>Previously, bzz (Alexander) wrote…</i></summary><blockquote>

Why is this dir and file needed, in such an unusual place?
</blockquote></details>

Sorry, it seems like if I tried to open `sbt` in the `native` folder by mistake. Deleted

ncordon:
___
*[src/main/scala/org/bblfsh/client/v2/BblfshClient.scala, line 153 at r5](https://reviewable.io/reviews/bblfsh/scala-client/116#-LnGPG-R0c7TyQug9o0y:-LnHrX9B5gJvCB0RJdz9:bfz7ga) ([raw file](https://github.com/bblfsh/scala-client/blob/72b1d5ad9e3f22b5b212d4f4c95e5d065604614d/src/main/scala/org/bblfsh/client/v2/BblfshClient.scala#L153)):*

Just to double check that I understood this correctly:

The issue is that the handle is small, so it doesn't add memory pressure to the JVM heap, but it pins a large buffer on the mirror side which keeps the process RSS high until _other_ JVM allocations trigger a GC?

I wonder if we could avoid doing this every time by only doing the GC when we know the buffer size is "fairly big" for some reasonable definition? (E.g., > 1MiB or something)
</blockquote></details>

Yeah, almost. I would add that since the buffer size is going to be much higher than the JVM ContextExt we generate (in JVM heap, and they only contain a `Long`) we generate with it, we would have to store a huge deal of `ContextExt` for the GC to wake up if we do not use this line. And we would be using a lot of RAM (I have peaked at 7 GiB) without the GC collecting the buffers :roll_eyes: Furthermore, if the buffer is not released, the Garbage Collector cannot release the native memory for `ContextExt` (because somehow interprets that `ContextExt` is using it, probably because of the call stack we generated and because it keeps a dependency tree inside to know what to deallocate first).

PR #118, opened 2019-08-19
Fixes README examples

ncordon:
Fixes the README example, which was not completely executable. 

It adds the syntax `client.filter` and `client.iterator` for `client : BblfshClient` to get closer to the `Python` API (up until now we had to call `BblfshClient.filter` and `BblfshClient.iterator` and that was another reason why the README example did not work).

It also changes the: `client.filter(resp.get, "//uast:Import")` (which was always empty iterator, at least for a Python pile) for `client.filter(resp.get, "//*uast:Import")`.

---
This change is [Reviewable](https://reviewable.io/reviews/bblfsh/scala-client/118)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,9 @@ target
 *~
 .idea
 src/libuast-native
+src/main/proto
+!src/main/proto/github.com/gogo/protobuf/gogoproto/gogo.proto
+src/main/resources
 .history
 Libuast.so
 build/*
--- a/README.md
+++ b/README.md
@@ -93,7 +93,7 @@ val resp = client.parse(filename, fileContent, Mode.SEMANTIC)
 println(resp.get)
 
 // Filtered response
-println(client.filter(resp.get, "//uast:Import"))
+println(client.filter(resp.get, "//uast:Identifier"))
 ```
 
 Command line:
@@ -105,7 +105,7 @@ java -jar build/bblfsh-client-assembly-*.jar -f <file.py>
 or if you want to use a XPath query:
 
 ```
-java -jar build/bblfsh-client-assembly-*.jar -f <file.py> -q "//uast:Import"
+java -jar build/bblfsh-client-assembly-*.jar -f <file.py> -q "//uast:Identifier"
 ```
 
 Please read the [Babelfish clients](https://doc.bblf.sh/user/language-clients.html)
--- a/README.md
+++ b/README.md
@@ -80,7 +80,7 @@ guide to learn more about how to use and deploy a bblfsh server, install languag
 API
 ```scala
 import scala.io.Source
-import org.bblfsh.client.BblfshClient
+import org.bblfsh.client.v2.BblfshClient, BblfshClient._
 import gopkg.in.bblfsh.sdk.v2.protocol.driver.Mode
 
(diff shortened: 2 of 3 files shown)


kuba--:
Do we need those `libuast`, `libscalauast` binaries?
If yes, why one is compiled as dynamic library (libscalauast.so) and another as static (libuast.a)?

ncordon:
No of course not, that's my bad. Good catch @kuba-- Fixed now!

ncordon:
___
*[.gitignore, line 7 at r1](https://reviewable.io/reviews/bblfsh/scala-client/118#-LmoN3BBDRwlblb81LnR:-LmoS5iv7E0FCb7_c4oz:bxyfig7) ([raw file](https://github.com/bblfsh/scala-client/blob/ed983f6c88419f832bbabdd8ad4ef201d93d2209/.gitignore#L7)):*
<details><summary><i>Previously, dennwc (Denys Smirnov) wrote…</i></summary><blockquote>

Wondering about this specific line. What is the reason for adding it?
</blockquote></details>

I want to ignore the files generated by `sbt assembly`, which are dumped in `src/main/proto`. I do not want to ignore `gogo.proto` file though, which needs to be there

ncordon:
___
*[README.md, line 96 at r1](https://reviewable.io/reviews/bblfsh/scala-client/118#-LmoMiYo6xBSmYoMAc2O:-LmoSNYG8TEXL8r4mdC-:b-jfqoqv) ([raw file](https://github.com/bblfsh/scala-client/blob/ed983f6c88419f832bbabdd8ad4ef201d93d2209/README.md#L96)):*
<details><summary><i>Previously, dennwc (Denys Smirnov) wrote…</i></summary><blockquote>

Hmm, why the `*` is used? It doesn't look right.

You probably need `//uast:RuntimeImport` instead.
</blockquote></details>

True. Leaving it as it was, since I think `RuntimeImport`s are only specific to Python files (which were my test files for this line), but should return something for files in any other language

ncordon:
___
*[README.md, line 96 at r1](https://reviewable.io/reviews/bblfsh/scala-client/118#-LmoMiYo6xBSmYoMAc2O:-LmoTJ453s4pA4L7-9kM:b-896fix) ([raw file](https://github.com/bblfsh/scala-client/blob/ed983f6c88419f832bbabdd8ad4ef201d93d2209/README.md#L96)):*

True. Leaving it as it was, since I think `RuntimeImport`s are only specific to Python files (which were my test files for this line), but should return something for files in any other language
</blockquote></details>

Done.

ncordon:
___
*[README.md, line 96 at r1](https://reviewable.io/reviews/bblfsh/scala-client/118#-LmoMiYo6xBSmYoMAc2O:-Lmt2J8hD3NEeU-IbHgw:b-896fja) ([raw file](https://github.com/bblfsh/scala-client/blob/ed983f6c88419f832bbabdd8ad4ef201d93d2209/README.md#L96)):*
<details><summary><i>Previously, dennwc (Denys Smirnov) wrote…</i></summary><blockquote>

Or we can change it to `//uast:Identifier` so it works for all languages.
</blockquote></details>

Done!

PR #121, opened 2019-08-22
Bumps libuast

### LINKED ISSUE #108: v2: expose new iteration orders
libuast provides multiple node iteration options: [pre, post, children, position, etc](https://github.com/bblfsh/libuast/blob/b69f7e5f7337db563e96f04dbcf91d6eb61271ef/src/uast.h#L64)

Right now we have a [4 orders exposed](https://github.com/bblfsh/scala-client/blob/fabffddbbc673a34b16637c88da5f8d660d71b22/src/main/scala/org/bblfsh/client/v2/BblfshClient.scala#L104) and this issue is about: 
 - bumping libuast version to the latest
 - making sure all the orders are exposed in JVM
 - adding more tests, covering other iterations orders [besides PreOrder](https://github.com/bblfsh/scala-client/blob/209838003bd42682178e1889dea7bfed44c8afb7/src/test/scala/org/bblfsh/client/v2/LibuastManagedIteratorTest.scala#L74)

ncordon:
* Bumps the libuast version.
* Exposes the tree orders from the C side: I have tried to lift the values as a value instead of as functions to reduce boilerplate code.
* Adds a `sbt-plugin` to ease development with JNI: [`sbt-jni`](https://github.com/jodersky/sbt-jni) and upadates `CONTRIBUTING.md` accordingly. With this plugin we can generate the C header files for the Scala code marked as `@native` automatically. Also note that some header files have been added without code, such as `org_bblfsh_client_v2_libuast_Libuast_UastIter__.h`. I prefer to leave those 2, 3 files there, since they are automatically generated by `sbt-jni` and it would add extra effort ask the developer to review those files manually if they execute `javah` and delete them if they do not have code.

Closes #108

---
This change is [Reviewable](https://reviewable.io/reviews/bblfsh/scala-client/121)


ncordon:
___
*[src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc, line 2 at r1](https://reviewable.io/reviews/bblfsh/scala-client/121#-Lmu5LkQ4avMA4y4k-pA:-Lmu5LkR-usHzqLdl6pu:b-a4ivc8) ([raw file](https://github.com/bblfsh/scala-client/blob/090bf1e2e4bfd5c4c30261f3afe277281a5667ea/src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc#L2)):*
> ```
> #include <map>
> ```

This line is necessary now, until we merge #116, which overwrites it. Former `libuast` used a `map` instead of `unordered_map`. With the version bump, we no longer include `map` with `libuast` and we need to do so manually here.

ncordon:
I have updated the `.travis.yml` to use xenial instead of trusty distribution. Reason: in trusty, the linker `apt-get` downloads seems to be old and generates a linking error

src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc
```diff
@@ -1,4 +1,5 @@
 #include <cassert>
+#include <map>
```
kuba--: is it (`map`) used somewhere, here?

ncordon:
___
*[src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc, line 2 at r2](https://reviewable.io/reviews/bblfsh/scala-client/121#-Lmu55yZXZQ71NVVSmGd-r2-2:-Lmy3Gly8aEl8k1V8_Z9:b-ayk0el) ([raw file](https://github.com/bblfsh/scala-client/blob/0fa67ca1420df54564b18d0564031bcaae40120b/src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc#L2)):*
<details><summary><i>Previously, kuba-- (Kuba Podgórski) wrote…</i></summary><blockquote>

is it (`map`) used somewhere, here?
</blockquote></details>

Yes, for the cache. Although I changed it to an `unordered_map` in `python-client` and @bzz did the same here, but in the PR #116 which is not merged yet. I explained why it is there in the Reviewable comments.

ncordon:
___
*[CONTRIBUTING.md, line 13 at r2](https://reviewable.io/reviews/bblfsh/scala-client/121#-Lmuzcv70cD3D9yCJm3I:-Lmy3p-l1IbjB-mDQihX:b-896fja) ([raw file](https://github.com/bblfsh/scala-client/blob/0fa67ca1420df54564b18d0564031bcaae40120b/CONTRIBUTING.md#L13)):*

s/should/must/

Also I recommend: s/C side/C/.
</blockquote></details>

Done!

src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc
bzz: Yes, it is used below as a KV cache.

PR #126, opened 2019-08-30
Adds UAST encoding options and improves tests

### LINKED ISSUE #107: v2: expose output UAST encoding options
libuast provides multiple encoding options: binary, yaml, JSON.

Right now we have a [binary on by default in ContextExt](https://github.com/bblfsh/scala-client/blob/fabffddbbc673a34b16637c88da5f8d660d71b22/src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc#L584) and [Context.encode()](https://github.com/bblfsh/scala-client/blob/fabffddbbc673a34b16637c88da5f8d660d71b22/src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc#L551), and this issue is about exposing other options though API surface, keeping binary a default.

### LINKED ISSUE #108: v2: expose new iteration orders

ncordon:
* Lifts the formats from `libuast` native side to the managed layer.
* Secures way of passing format parameters and tree orders for iterators (we do not accept an incorrect number anymore).
* Improves tests for iterators: know table-driven, as @bzz requested and closer to `python-client` ones.
* Adds tests for encoding / decoding from the different formats.

Closes #107, closes #108

This change is [Reviewable](https://reviewable.io/reviews/bblfsh/scala-client/126)


ncordon:
___
*[src/main/scala/org/bblfsh/client/v2/BblfshClient.scala, line 157 at r1](https://reviewable.io/reviews/bblfsh/scala-client/126#-LnYAfsx21xJbICEBI5n:-LnYNaRoCpW_bJXKavtO:b-hrt91s) ([raw file](https://github.com/bblfsh/scala-client/blob/98f2ea3fd637632352baf74e9c102ea71cfe7d23/src/main/scala/org/bblfsh/client/v2/BblfshClient.scala#L157)):*

As written I think this is a breaking change to the API. Should we provide a default argument?

(I believe only DP team may be using this right now, but I'd really like to avoid breaking them without a good reason)

This also applies below.
</blockquote></details>

No no, I have not touched the API at all. The syntax `buf.decode()` was not provided by these functions, but by an `implicit` [here](https://github.com/ncordon/scala-client/blob/98f2ea3fd637632352baf74e9c102ea71cfe7d23/src/main/scala/org/bblfsh/client/v2/BblfshClient.scala#L162). `implicits` lie in memory and whenever I try to use `buf.decode()` Scala says: okay I do not know how to call `decode` on a `ByteBuffer` explicitly, do I have any `implicits` which tells me how to do it in my runtime? Here is some documentation for them: https://docs.scala-lang.org/overviews/core/implicit-classes.html

What I do not recommend is having default parameters in Scala, because they are not compatible with Java, as documented [here](https://docs.scala-lang.org/tour/default-parameter-values.html). Of course we can use them if you prefer to do it. But that is why I have duplicated the `decode` methods for example into:

```
def decode(buf, format)
def decode(but) = decode(buf, UastBinary)
```

ncordon:
___
*[src/main/scala/org/bblfsh/client/v2/package.scala, line 14 at r1](https://reviewable.io/reviews/bblfsh/scala-client/126#-LnYBhYg3H-f1dCfh-_g:-LnYSeWKBEt8QxESbtmq:b-abrhrq) ([raw file](https://github.com/bblfsh/scala-client/blob/98f2ea3fd637632352baf74e9c102ea71cfe7d23/src/main/scala/org/bblfsh/client/v2/package.scala#L14)):*

I'm not sure _why_ we would want to pun away the types here. Is this needed for some other interface we're relying on?
</blockquote></details>

I am only exposing the `implicit` conversions for `TreeOrder`s and `UastFormat`s. Let me explain a little bit. I did not want to break the API, because what we had until now was for example: `def iterator(..., order: Int)` or `def decode(..., fmt:Int)`. Obviously if I introduced an integer which was not supported by `libuast` everything blew up. So I have turned those methods into `iterator(..., order: TreeOrder)` and `decode(..., fmt: UastFormat)`, which are type-safe and would fail if I do not introduce a correct order / format. That being said, I wanted the old syntax `iterator(..., 0)` to work to avoid breaking the API, so I have to give Scala an implicit way of converting an integer to a correct `TreeOrder` (we use a default order warning the user when they introduce something invalid, like `iterator(..., -1)`)

ncordon:
___
*[src/main/scala/org/bblfsh/client/v2/package.scala, line 14 at r1](https://reviewable.io/reviews/bblfsh/scala-client/126#-LnYBhYg3H-f1dCfh-_g:-LnYUMmx-kl2JiSzVAJd:b-xva271) ([raw file](https://github.com/bblfsh/scala-client/blob/98f2ea3fd637632352baf74e9c102ea71cfe7d23/src/main/scala/org/bblfsh/client/v2/package.scala#L14)):*

I am only exposing the `implicit` conversions for `TreeOrder`s and `UastFormat`s. Let me explain a little bit. I did not want to break the API, because what we had until now was for example: `def iterator(..., order: Int)` or `def decode(..., fmt:Int)`. Obviously if I introduced an integer which was not supported by `libuast` everything blew up. So I have turned those methods into `iterator(..., order: TreeOrder)` and `decode(..., fmt: UastFormat)`, which are type-safe and would fail if I do not introduce a correct order / format. That being said, I wanted the old syntax `iterator(..., 0)` to work to avoid breaking the API, so I have to give Scala an implicit way of converting an integer to a correct `TreeOrder` (we use a default order warning the user when they introduce something invalid, like `iterator(..., -1)`)
</blockquote></details>

I would rather erase this of course, but then we delete the backward compatibility with the old syntax passing numbers, and I prefer to have these things here. They are here because they would be in scope for the whole package, since `treeOrder`s and `uastFormat`s are used in several files from the library and I want the old syntax to work everywhere inside the package

ncordon:
___
*[src/main/scala/org/bblfsh/client/v2/BblfshClient.scala, line 157 at r1](https://reviewable.io/reviews/bblfsh/scala-client/126#-LnYAfsx21xJbICEBI5n:-LnYYFD_Cg9HYmmDAgkX:bga60bb) ([raw file](https://github.com/bblfsh/scala-client/blob/98f2ea3fd637632352baf74e9c102ea71cfe7d23/src/main/scala/org/bblfsh/client/v2/BblfshClient.scala#L157)):*

Understood, thanks for the explanation!

Given that this `decode` is to satisfy the interface, and isn't called directly by the client, I agree it doesn't make sense to add in defaults.

(As an aside: Implicit classes are a great example of a feature that really helps the author of the code, at the expense of the reader. I wish language designers would think about both sides more carefully.)
</blockquote></details>

Yes, totally agree, when I started programming in Scala it was driving me mad constantly not knowing where the definitions for thing were because of the `implicit`s stuff

ncordon:
Just to clarify, tests are going to fail because in the file `BblfshClientParseTest.scala` I have included a test to check  that `encode`is the inverse method to `decode`, which right now is not true for `YAML` format. As explained, a list of roles like this:

```
Visibility, World, Declaration, Type
```

would be encoded as:

```
Declaration, Type, Visibility, World
```

due to this [line](https://github.com/bblfsh/sdk/blob/c7df66261dcabae8ddd71c6a68a4391212459e42/uast/uastyaml/uastyaml.go#L108) which is code called by `libuast`. We are currently discussing about either sorting in decoding too or dropping that line.

Issue #127, opened 2019-09-10
Why JUInts?


ncordon:
Is there any strong reason to use [`JUints`](https://github.com/bblfsh/scala-client/blob/c4c15bbf848d17538f42ea09b370b1ece32795f8/src/main/scala/org/bblfsh/client/v2/NodeExt.scala#L153) when Scala does not support the `Unsigned Int` type and [`python-client`](https://github.com/bblfsh/python-client/blob/f4ddc558d49f5fb3e93f73b88d98548adc40ebce/bblfsh/pyuast.cc#L455) does not support them either?

Well, in fact in the `python-client` we have [ways of casting](https://github.com/bblfsh/python-client/blob/f4ddc558d49f5fb3e93f73b88d98548adc40ebce/bblfsh/pyuast.cc#L517) to `uint64` in the native part, but they are never going to get called, since the type of node, as linked before, is not going to ever return `NODE_UINT`

bzz:
Good point. Honestly, I do not think we really need it. 

It was just added as an attempt to implement the interface that we have on the native side with the conversion in the managed code.

I think it's safe to drop it and document that behavior is similar to python-client.

ncordon:
This can be closed, since having `JUint`s seems to be protecting against losing bytes in `encoding ∘decoding` operations. The same cannot be said about the `python-client`, so will open an issue there.

PR #128, opened 2019-09-11
Adds value based filtering

### LINKED ISSUE #110: v2: expose filtering for values
Make it possible to query for individual value node types again and document it in README:

```scala
// Filtered using XPath functions returning value types
client.filterBool(resp.get, "boolean(//*[@strtOffset or @endOffset])")
client.filterString(resp.get, "name(//*[1])")
client.filterNumber(resp.get, "count(//*)")
```

ncordon:
Exposes, in `BblfshClient`:

* `filterInt` -> returns only nodes for the query which are `JInt`s
* `filterUint` -> analogously, but with `JUint`s
* `filterFloat` -> same for `JFloat`s
* `filterString` -> same for `JString`s
* `filterBool` -> same for `JBool`s

Note that we may want to have a clear answer for #127 before merging this, because in the python-client (where we do not have `uint`s) the analogous functionality is to call:

```python
it = filter(resp.get, boolean(//*))

for node:
  # This may fail if any of the nodes we are filtering is not of boolean type
  node.get_bool()
```

Also in the issue we were talking about `filterNum`, but unless we have a type that groups `JFloat`, `JInt` and `JUint`, we cannot do that. Moreover, in the `python-client`, the filtering of `float`s is separated from the filtering of `int`s.

Closes #110

---
This change is [Reviewable](https://reviewable.io/reviews/bblfsh/scala-client/128)

### DIFF
--- a/src/main/scala/org/bblfsh/client/v2/BblfshClient.scala
+++ b/src/main/scala/org/bblfsh/client/v2/BblfshClient.scala
@@ -240,9 +240,10 @@ object BblfshClient {
     // do not dispose the context, iterator steals it
   }
 
-  // Method to filter managed nodes based on type
-  // The aim is to not return a JNode but a specialized subclass of JNode, as JBool,
-  // for example, when T = JBool
+  /** Method to filter managed nodes based on type
+    * The aim is to not return a JNode but a specialized subclass of JNode, as JBool,
+    * for example, when T = JBool
+    */
   private def filterOnType[T : ClassTag](node: JNode, query: String): Iterator[T] = {
     val it = filter(node, query)
     // Filter only the nodes that are of type T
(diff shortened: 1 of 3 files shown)


ncordon:
Good question, I asked myself the same thing. Probably @bzz can clarify that for us when he returns.

src/main/scala/org/bblfsh/client/v2/BblfshClient.scala
```diff
@@ -218,21 +223,40 @@ object BblfshClient {
...
+  /** Factory method for iterator over a managed node, filtered by XPath query */
+  def filter(node: JNode, query: String): Libuast.UastIter = Libuast.synchronized {
     val ctx = Context()
     ctx.filter(query, node)
     // do not dispose the context, iterator steals it
   }
 
+  // Method to filter managed nodes based on type
+  // The aim is to not return a JNode but a specialized subclass of JNode, as JBool,
+  // for example, when T = JBool
```
bzz: I suggest converting these comments to a ScalaDoc.

ncordon: Done!

PR #133, opened 2019-09-17
Adds gRPC timeout parameter and bumps default timeout

### LINKED ISSUE #100: v2: expose custom gRPC connection timeouts
The gRPC connection timeouts are important for parsing big files (see #79) but with intial migration to protocol.v2 we lost previously `.parseWithTimeout()`.

This task is about enabling/documenting timeouts at v2 again.

Initial discussion at code review:
 - [feedback#1](https://github.com/bblfsh/client-scala/pull/84#discussion_r288271268)
 - [feedback#2](https://github.com/bblfsh/client-scala/pull/84#discussion_r288357335)

ncordon:
Closes #100 

`timeout` was not working with the migration to `v2`. And based on the feedback:
- @creachadair proposed to increase default timeout to 1 or 2 min. Done!. But note we should probably do the same in the `python-client` then, which if no timeout is passed, I think it will not timeout.
- @dennwc proposed to unify the API for the `BblfshClient.parse` method, adding default parameters. Although I am in favour of that, adding default parameters will break the API for a Java user of this package: source [here](https://docs.scala-lang.org/tour/default-parameter-values.html). 

For example a method:

```
def Point(val x: Double = 0)
```

could be called as `Point()` in Scala, but not in Java.

If we unify `parse` methods we have to bump the major version of the package. So wdyt?

---
This change is [Reviewable](https://reviewable.io/reviews/bblfsh/scala-client/133)

### DIFF
--- a/src/main/scala/org/bblfsh/client/v2/BblfshClient.scala
+++ b/src/main/scala/org/bblfsh/client/v2/BblfshClient.scala
@@ -5,11 +5,13 @@ import java.nio.ByteBuffer
 import com.google.protobuf.ByteString
 import gopkg.in.bblfsh.sdk.v2.protocol.driver._
 import io.grpc.ManagedChannelBuilder
+import java.util.concurrent.TimeUnit
 import org.bblfsh.client.v2.libuast.Libuast
 
 
 class BblfshClient(host: String, port: Int, maxMsgSize: Int) {
-  private val DEFAULT_TIMEOUT_SEC = 5
+  // 1 minute default timeout
+  val DEFAULT_TIMEOUT_SEC = 60
 
   private val channel = ManagedChannelBuilder
     .forAddress(host, port)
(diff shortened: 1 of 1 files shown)


ncordon:
Well, not really. However, as far as my understanding goes, there is no such thing as Scala package, Kotlin package or Java packages nowadays. There are JVM packages, and you can import them from whichever language compiles to JVM bytecode. Maybe I am wrong about this.

Anyway, do you propose any way of merging the following methods?

```
def parseWithOptions(name: String, content: String, lang: String, timeout: Long, mode: Mode)
def parse(name: String, content: String, lang: String = "")
def parse(name: String, content: String, mode: Mode)
def parseWithTimeout(name: String, content: String, timeout: Long, lang: String = "")
```

All of them are public, so I did not see any obvious way of merging them without breaking the API backwards compatibility.

PR #134, opened 2019-09-20
Refactor CI

### LINKED ISSUE #87: Improve the CI
Potential action items:

 - make release job more modular https://github.com/bblfsh/client-scala/pull/84#discussion_r289942135
 - do not install all native dependencies (presumably, some are only necessary for cross-compilation)
 - do not install all recommended drivers for `bblfshd`
 - check if building in separate job on OS X vm is faster then cross-compile and add it to normal CI flow

### LINKED ISSUE #88: Restructure build
Right now, build consists of:

 - native part
 - Scala part

all being handled by `sbt`. As [noted](https://github.com/bblfsh/client-scala/pull/84#discussion_r289940924) it would be a good idea to have a Makefile that CI uses, that takes care of native part and calls into the `./sbt`.

### LINKED ISSUE #102: Windows platform support
It looks like the current scala-client doesn't properly handle Windows. The same code works on Ubuntu 18.04 but causes an NPE at https://github.com/bblfsh/scala-client/blob/master/src/main/scala/org/bblfsh/client/v2/libuast/Libuast.scala#L62 on Windows.

It seems this is due to the scala-client not embedding the `libuast.dll` and loading this instead of `libscalauast.so`. Is this correct? If so, shouldn't the effort to get scala-client working with Windows be fairly low? I'm only seeing small changes to `Libuast.scala` and `build.sbt` being necessary. Am I missing something?

ncordon:
- [x] Builds for OSX in that operating system instead of cross compile. This cleans the pair of hacks we had in the `.travis.yml` script.
- [x] Adds support for Windows, adding a new job in Travis that builds the `dll`
- [x] Pushes static libraries generated for Linux, OSX and Windows to GH releases
- [x] Creates new job that gathers those 3 libraries and builds the `scala-client` with them embedded. Downside: that increases the size of the package.
- [x] Simplifies some of the build process using a ~`makefile`~ bash script to split fetching of dependencies from build of the project. README updated with new way of fetching dependencies with the script and compiling the project.

The process in Travis would look something alike:

```
Build and test in Linux ----> [if tag] Push .so to GH releases  ----------|
                                                                          |
Build in OSX ---------------> [if tag] Push .dylib to GH releases  --------------> [if tag] Download so,dylib,dll,    
                                                                          |                 build, push to Maven
Build in Windows -----------> [if tag] Push .dll to GH releases ----------|
```

Closes #87, closes #88, closes #102 

---
This change is [Reviewable](https://reviewable.io/reviews/bblfsh/scala-client/134)


ncordon:
___
*[src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc, line 844 at r1](https://reviewable.io/reviews/bblfsh/scala-client/134#-LpbpoF1E1NYXt7B85t4:-LpbpoF1E1NYXt7B85t5:b-knj05e) ([raw file](https://github.com/bblfsh/scala-client/blob/8d2f15c97aa2b6fc14262ce7592d287aa4a3d70d/src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc#L844)):*
> ```cpp
> Java_org_bblfsh_client_v2_Context_00024_create(JNIEnv *env, jobject self) {
>   Context *c = new Context();
>   return (jlong) c;
> ```

I am concerned about this. I mean, there is no change in this PR really, I am just making sure to cast the pointer to `jlong` and not just `long`, which is what we did before. But `jlong` in my Linux is a `long` (note this is what we had before), and in the Windows setting I have tested, a `long long`. What would happen if `Context* c` is assigned a `long long` (64 bits) address in memory? Are we casting it to a `long` (32 bits) in the case of Linux? That probably would result in a SIGSEV if we try to access a wrong address

ncordon:
___
*[build.sh, line 132 at r1](https://reviewable.io/reviews/bblfsh/scala-client/134#-LpbqXzC32q2mfIVg3kz:-LpbqXzC32q2mfIVg3l-:bh9lcyi) ([raw file](https://github.com/bblfsh/scala-client/blob/8d2f15c97aa2b6fc14262ce7592d287aa4a3d70d/build.sh#L132)):*
> ```Shell
> }
> 
> for arg in "$@"; do
> ```

Suggestions about better ways of doing the argument parsing here? I have tested `getopt` command, but did not managed to make it work correctly

ncordon:
___
*[build.sh, line 132 at r1](https://reviewable.io/reviews/bblfsh/scala-client/134#-LpbqXzC32q2mfIVg3kz:-Lpbqgi74gFZhFe075V3:b-tgjp0a) ([raw file](https://github.com/bblfsh/scala-client/blob/8d2f15c97aa2b6fc14262ce7592d287aa4a3d70d/build.sh#L132)):*

Suggestions about better ways of doing the argument parsing here? I have tested `getopt` command, but did not managed to make it work correctly
</blockquote></details>

Desirable thing is I have an autocompletion pressing TAB for the commands available in the script (do not know if that can be done), and an order of arguments that does not matter (it should only matter the fed commands, not the order we feed them to the script)

ncordon:
@bblfsh/maintainers ready for review :smile:

ncordon:
___
*[.travis.yml, line 71 at r3](https://reviewable.io/reviews/bblfsh/scala-client/134#-Lpd_PB9CDh5gmQBagid:-Lpm5sUo0ENuw2NrmOcd:bbu5pbm) ([raw file](https://github.com/bblfsh/scala-client/blob/3f907d01a793e6a8a3db06212b98afee4f8e19d9/.travis.yml#L71)):*

Is `choco` part of the base image, or do we need to install it?

Separately: Can you please add a comment about why this particular version (and/or when it is safe to switch it)? It doesn't need to be exhaustively detailed, just a breadcrumb for someone doing updates later.
</blockquote></details>

`choco` is part of the Windows image Travis provides. The version is just for convenience. We need JDK 8, but since we need to export the JAVA_PATH after that we need to choose a revision of the v8 (the path depends on it). It may be safe to update as long as someone gets the `JAVA_PATH` updated correctly also.

ncordon:
___
*[.travis.yml, line 71 at r3](https://reviewable.io/reviews/bblfsh/scala-client/134#-Lpd_PB9CDh5gmQBagid:-Lpm6ItbAdjy723-8L0T:boagh9o) ([raw file](https://github.com/bblfsh/scala-client/blob/3f907d01a793e6a8a3db06212b98afee4f8e19d9/.travis.yml#L71)):*

`choco` is part of the Windows image Travis provides. The version is just for convenience. We need JDK 8, but since we need to export the JAVA_PATH after that we need to choose a revision of the v8 (the path depends on it). It may be safe to update as long as someone gets the `JAVA_PATH` updated correctly also.
</blockquote></details>

Will add the comment of course

ncordon:
___
*[.travis.yml, line 81 at r3](https://reviewable.io/reviews/bblfsh/scala-client/134#-Lpd_l7qBm1FKWcO4xTS:-Lpm7ipzEoyUGJTbRUit:bbf8qff) ([raw file](https://github.com/bblfsh/scala-client/blob/3f907d01a793e6a8a3db06212b98afee4f8e19d9/.travis.yml#L81)):*
> set -e

Nope, but I should add it probably, shouldn't I?

ncordon:
___
*[build.sh, line 2 at r3](https://reviewable.io/reviews/bblfsh/scala-client/134#-LpdatTs2aYDsy89G1y6:-Lpm7wch09eFYuVpsxbB:b-flmbi) ([raw file](https://github.com/bblfsh/scala-client/blob/3f907d01a793e6a8a3db06212b98afee4f8e19d9/build.sh#L2)):*

Please add a comment below the shebang line, to summarize what this script does.

My main high-level comment is that this script isn't consistent about error-checking. Some of the functions do use `&&`, but not all—and if something goes wrong with (say) fetching the SDK the world will be in an unknown state. Consider maybe adding `set -e` so that execution errors will terminate the script early rather than cascading. That will also let you drop all the `&&` cascades and maybe that will be easier to read.
</blockquote></details>

If I add `set -e` at the beginning is enough? Or do I have to add it at the beginning of each function?

ncordon:
___
*[build.sh, line 49 at r3](https://reviewable.io/reviews/bblfsh/scala-client/134#-Lpdb8-V9IdvT3SP7iNn:-Lpm8GSW6uaix4bRoSvX:bebrt1w) ([raw file](https://github.com/bblfsh/scala-client/blob/3f907d01a793e6a8a3db06212b98afee4f8e19d9/build.sh#L49)):*

Please quote and terminate arguments: `rm -rf -- "${UNZIP_DIR}"
</blockquote></details>

Will do

ncordon:
___
*[build.sh, line 68 at r3](https://reviewable.io/reviews/bblfsh/scala-client/134#-LpdbF2f5GLWvisvtSUC:-Lpm8INj9xNCTTpNgiJp:b4putsz) ([raw file](https://github.com/bblfsh/scala-client/blob/3f907d01a793e6a8a3db06212b98afee4f8e19d9/build.sh#L68)):*

Maybe put the diagnostics to stderr? 

```bash
echo "[sdk] Done unpacking SDK" 1>&2
```
</blockquote></details>

:ok_hand:

ncordon:
___
*[build.sh, line 73 at r3](https://reviewable.io/reviews/bblfsh/scala-client/134#-LpdbSmf3f5jpyKETYk9:-Lpm8Lwe26l4u9QaKcuU:bjqny46) ([raw file](https://github.com/bblfsh/scala-client/blob/3f907d01a793e6a8a3db06212b98afee4f8e19d9/build.sh#L73)):*

I'm not sure why this is here? Is this just for diagnostic purposes? 

(If so: Maybe redirect to stderr? Possibly also filter plain files: `find src/main/resources -type f 1>&2`)
</blockquote></details>

This was copied from the Scala build script. It just lists the folder where the library should have been generated. I tried to change it for a `ls -R /src/main/resources` but the output of the `find` command is much clearer.

ncordon:
___
*[build.sh, line 149 at r3](https://reviewable.io/reviews/bblfsh/scala-client/134#-LpdcaCm-FqoHjaU05Ko:-Lpm8vdT9x5IhF5WhoRb:b4putsz) ([raw file](https://github.com/bblfsh/scala-client/blob/3f907d01a793e6a8a3db06212b98afee4f8e19d9/build.sh#L149)):*

(minor nit) probably "invalid" or "unsupported" rather than "wrong"
</blockquote></details>

:ok_hand:

ncordon:
___
*[src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc, line 844 at r1](https://reviewable.io/reviews/bblfsh/scala-client/134#-LpbpoF1E1NYXt7B85t4:-LpmWd5m7e2nCpuFf4MF:b-7fryki) ([raw file](https://github.com/bblfsh/scala-client/blob/8d2f15c97aa2b6fc14262ce7592d287aa4a3d70d/src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc#L844)):*
> static_cast<jlong>(c)

`static_cast cannot convert from ‘{anonymous}::Context*’ to ‘jlong’ {aka ‘long int’}`

The only material I have found about this is in [Google's JNI tips](https://developer.android.com/training/articles/perf-jni) which says:

> To support architectures that use 64-bit pointers, use a long field rather than an int when storing a pointer to a native structure in a Java field.

which clearly is not working for 64 bits (compiler complains). I have also found [this blog post](https://hypertolosana.wordpress.com/2015/04/30/android-jni-c-pointers-to-java/) saying that doing `jlong(c)` (which is equivalent to what we have done should work, whereas `(long) c` does not). But I would like some sort of official docs confirmation

ncordon:
I can confirm this seems to work both from a Git Bash and a Powershell in Windows

ncordon:
Friendly ping @creachadair. All the comments have been addressed (I think)

ncordon:
___
*[build.sh, line 2 at r3](https://reviewable.io/reviews/bblfsh/scala-client/134#-LpdatTs2aYDsy89G1y6:-LqHUOLeEfLQCQEi47SD:b3vb5hs) ([raw file](https://github.com/bblfsh/scala-client/blob/3f907d01a793e6a8a3db06212b98afee4f8e19d9/build.sh#L2)):*

You don't need to add it to each function, setting it at the top level affects the whole shell (though I think it does not propagate to subshells).
</blockquote></details>

Are you sure? I read this in `man sed`:

```
         Application writers should avoid relying on set −e within functions. For example, in the following script:

           set -e
           start() {
               some_server
               echo some_server started successfully
           }
           start || echo >&2 some_server failed

       the −e setting is ignored within the function body (because the function is a  command  in  an  AND-OR  list
       other  than the last). Therefore, if some_server fails, the function carries on to echo "some_serverstarted‐
       successfully", and the exit status of the function is zero (which means "some_serverfailed" is not output).
```

ncordon:
___
*[build.sh, line 73 at r3](https://reviewable.io/reviews/bblfsh/scala-client/134#-LpdbSmf3f5jpyKETYk9:-LqHVDbx-4kxT-lNJfwB:b-896fja) ([raw file](https://github.com/bblfsh/scala-client/blob/3f907d01a793e6a8a3db06212b98afee4f8e19d9/build.sh#L73)):*

Having it as `find` is fine by me, I just wasn't sure why. It sounds like it's just for diagnostics, in which case I'd suggest including the redirect to stderr so that all the log output stays together.
</blockquote></details>

Done!

ncordon:
___
*[.travis.yml, line 81 at r3](https://reviewable.io/reviews/bblfsh/scala-client/134#-Lpd_l7qBm1FKWcO4xTS:-LqHXRQ9C2Fj4Mxp-7_9:b-bzoh5e) ([raw file](https://github.com/bblfsh/scala-client/blob/3f907d01a793e6a8a3db06212b98afee4f8e19d9/.travis.yml#L81)):*

As a rule of thumb I prefer to do it for setup scripts, where otherwise ignoring an an early failure can make it difficult to debug an indirect error later. I wouldn't bother for diagnostic or cleanup scripts, where a failure doesn't affect the correctness of the results. 
</blockquote></details>

Problem with the `before_script` section I think is that even if we get an error there I think the `script` section would execute anyway. Should I merge `before_script` and `script`?

ncordon:
___
*[build.sh, line 2 at r3](https://reviewable.io/reviews/bblfsh/scala-client/134#-LpdatTs2aYDsy89G1y6:-LqHXvaO3xGRpvqW-CcW:bj0el4a) ([raw file](https://github.com/bblfsh/scala-client/blob/3f907d01a793e6a8a3db06212b98afee4f8e19d9/build.sh#L2)):*

That advice is good, because `awk` doesn't know what shell it's using, and POSIX does not specify IIRC. But here we explicitly specify Bash, which I _think_  does the right thing at least in versions we are likely to see anywhere in the wild. 

I believe it is harmless to add it explicitly in each function, however, at least in Bash. I certainly don't object to doing both if you want to play it safe.
</blockquote></details>

Added to the header of the file also!

ncordon:
___
*[src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc, line 844 at r1](https://reviewable.io/reviews/bblfsh/scala-client/134#-LpbpoF1E1NYXt7B85t4:-LqHc8pwCY96wVDm84lU:bg4p086) ([raw file](https://github.com/bblfsh/scala-client/blob/8d2f15c97aa2b6fc14262ce7592d287aa4a3d70d/src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc#L844)):*

Ah, I overlooked that we are starting from a pointer, so what we would really want in this case is `reinterpret_cast` not `static_cast`.  But that's basically the same effect apart from notation.
</blockquote></details>

Changed to `reinterpret_cast` now

ncordon:
___
*[.travis.yml, line 81 at r3](https://reviewable.io/reviews/bblfsh/scala-client/134#-Lpd_l7qBm1FKWcO4xTS:-LqHiYwz1THCdSfV1FFy:bbxznds) ([raw file](https://github.com/bblfsh/scala-client/blob/3f907d01a793e6a8a3db06212b98afee4f8e19d9/.travis.yml#L81)):*

Hmm. If indeed an error in `before_script` doesn't prevent `script` from running, that would probably make sense. It seems like if we don't succeed in fetching these various assets, there's no way the rest of the build can succeed. (Or if it could: Maybe we can just delete these assets entirely? :grinning:)
</blockquote></details>

I have moved the download of those assets to script :)

PR #138, opened 2019-10-04
Adds cache for methods and classes to the JNI

### LINKED ISSUE #114: v2: add a cache to JNI part
On the JNI side, we can cache a lot of things as global references instead of calling JVM every time:
 - classes, instead of `env->FindClass`
 - methodIds, instead of `env->GetMethodID` for things in hot `checkJvmException`

This can be done in `JNI_OnLoad` and all global reference must be deleted e.g in `JNI_OnUnLoad`.

ncordon:
Caches methods and classes in the JNI. I have also tried to cache `FieldIDs`, but is an awful idea, since right now the code looks like:

```cpp
jfieldID FieldID(JNIEnv *env, jobject obj, const char *field, const char *typeSignature) {
  jclass cls = env->GetObjectClass(obj);
  jfieldID fId = env->GetFieldID(cls, field, typeSignature);
  return fId;
}
``` 

So I need a cache `jclass -> jfieldID`. A `jclass` is a subclass of `jobject`, and as we encountered before in #113 `jobject`s cannot be used as key for an `unordered_map` unless we modify the hash funcion to call the managed `IsSameObject` method or `hashCode`. And I think it's just not worthy it in that case in terms of speed up of the code for `FieldID`s

Closes #114 

---
This change is [Reviewable](https://reviewable.io/reviews/bblfsh/scala-client/138)


ncordon:
___
*[src/main/native/jni_utils.h, line 58 at r1](https://reviewable.io/reviews/bblfsh/scala-client/138#-LqMnJ-A4mWlwPm0bUpY:-LqN6t6523wCPxiN8m5k:b-a2ujfy) ([raw file](https://github.com/bblfsh/scala-client/blob/e58a8b71e22b1ea16bf2e0abc3a968d8b640b62c/src/main/native/jni_utils.h#L58)):*

Minor nit: I think this operation is usually called `clear` or `reset` rather than `clean`, but YMMV.
</blockquote></details>

Changed!

ncordon:
___
*[src/main/native/jni_utils.cc, line 74 at r1](https://reviewable.io/reviews/bblfsh/scala-client/138#-LqMnTAW1pXGRW-qYLpp:-LqN7kEl2-CS4x7X5cJs:b5ey328) ([raw file](https://github.com/bblfsh/scala-client/blob/e58a8b71e22b1ea16bf2e0abc3a968d8b640b62c/src/main/native/jni_utils.cc#L74)):*

Minor thing: Even though they're not required in this case, please consider using explicit brackets anyway, here and below.
</blockquote></details>

Done! Thanks :)

ncordon:
___
*[src/main/native/jni_utils.cc, line 85 at r1](https://reviewable.io/reviews/bblfsh/scala-client/138#-LqMnipK4ZYnzErJXYI0:-LqN87-W6p-ag8yefk77:bmk2znb) ([raw file](https://github.com/bblfsh/scala-client/blob/e58a8b71e22b1ea16bf2e0abc3a968d8b640b62c/src/main/native/jni_utils.cc#L85)):*

Should we return early in this case, to avoid populating the cache with a null reference?
</blockquote></details>

In the [documentation](https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/functions.html#FindClass) it says it would throw an exception in case the name is not from a valid class, and `checkJVMException` would catch that and return from the native code.  So we are safe 8-)

ncordon:
___
*[src/main/native/jni_utils.cc, line 85 at r1](https://reviewable.io/reviews/bblfsh/scala-client/138#-LqMnipK4ZYnzErJXYI0:-LqNW3mu9FfAw4CQNLUV:bvnibdr) ([raw file](https://github.com/bblfsh/scala-client/blob/e58a8b71e22b1ea16bf2e0abc3a968d8b640b62c/src/main/native/jni_utils.cc#L85)):*

Right—but as far as I can see `checkJVMException` doesn't actually throw a _native_ exception, so after signaling the exception in the environment, the rest of this function will execute and poke the (invalid) `localClassRef` into the cache before yielding back to the runtime.

If that's correct, then wouldn't this leave the cache in a bad state? If the user thread handles the exception and then resumes, won't the next cache lookup on this name obtain what looks (incorrectly) like a hit?

(It is quite likely I overlooked some aspect of the control flow here, I'm just trying to make sure we leave things in a good state)
</blockquote></details>

Oh no, you are absolutely correct. I figured `env->ThrowNew` returned inmediately

ncordon:
___
*[src/main/native/jni_utils.cc, line 85 at r1](https://reviewable.io/reviews/bblfsh/scala-client/138#-LqMnipK4ZYnzErJXYI0:-LqNX0wAB1zVzLdOPvsF:bhqqzh5) ([raw file](https://github.com/bblfsh/scala-client/blob/e58a8b71e22b1ea16bf2e0abc3a968d8b640b62c/src/main/native/jni_utils.cc#L85)):*

Oh no, you are absolutely correct. I figured `env->ThrowNew` returned inmediately
</blockquote></details>

*immediately

ncordon:
___
*[src/main/native/jni_utils.cc, line 85 at r1](https://reviewable.io/reviews/bblfsh/scala-client/138#-LqMnipK4ZYnzErJXYI0:-Lqez8gk8dw3a2HWxf3S:b-uua14r) ([raw file](https://github.com/bblfsh/scala-client/blob/e58a8b71e22b1ea16bf2e0abc3a968d8b640b62c/src/main/native/jni_utils.cc#L85)):*

This is an area of JNI that I'm not super familiar with, but I think the exception doesn't trigger till the extension returns. 
</blockquote></details>

Yes, as said you are right :) I am going to review other areas of the code just in case we/I have incorrectly used it somewhere else

ncordon:
___
*[src/main/native/jni_utils.cc, line 191 at r3](https://reviewable.io/reviews/bblfsh/scala-client/138#-LqkvkzN3h9sDmkb0xkU:-LqkvkzN3h9sDmkb0xkV:b-2a9yk3) ([raw file](https://github.com/bblfsh/scala-client/blob/2ea8ad4deaae166cf06084b3d389b2433c1d703e/src/main/native/jni_utils.cc#L191)):*
> ```cpp
>     checkJvmException(std::string("failed to get field ID for field name '")
>                           .append(name)
>                           .append("' with signature '")
> ```

This has been changed since it threw the exception manually, when we have a wrapper for that

ncordon:
___
*[src/main/native/jni_utils.cc, line 212 at r3](https://reviewable.io/reviews/bblfsh/scala-client/138#-LqkwAML6SEAM0w_QgqP:-LqkwAML6SEAM0w_QgqQ:b-fihtsw) ([raw file](https://github.com/bblfsh/scala-client/blob/2ea8ad4deaae166cf06084b3d389b2433c1d703e/src/main/native/jni_utils.cc#L212)):*
> ```cpp
>                           .append(signature)
>                           .append("'"));
>     return -1;
> ```

Likewise

PR #139, opened 2019-10-04
Reduces size of native shared libraries

### LINKED ISSUE #85: Investigate released artifact size increase
Investigate and document JAR file increase -  v1.11.0 jar [published on Maven](https://oss.sonatype.org/#nexus-search;quick~bblfsh) is 1.31 MB while `bblfsh-client-2.0.0-SNAPSHOT.jar` from #83 is 8.0 MB.

ncordon:
Uses `g++ -s` to strip all debug symbols and ignores `libuast.a` and `libuast.so` in the resulting `.jar` file. Note the contents of `libuast` folder were no needed in the `.jar` since we have already created `libscalauast.so` using the static library.

This should reduce the size of the `.jar`s even now that we have three flavours of `libscalauast` dynamic library (for Linux, macOS and Windows)

Closes #85 

---
This change is [Reviewable](https://reviewable.io/reviews/bblfsh/scala-client/139)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -10,7 +10,7 @@ This section has more extended documentation for a brave developer willing to
 contribute by diving into and coding / debugging for the native glue code in `src/main/native`.
 
 ## Generate C headers for native code
-A method must be marked as `@native` in the Scala part for us to be able to code it natively in C. 
+A method must be marked as `@native` in the Scala part for us to be able to code it natively in C.
 Example:
 
 ```scala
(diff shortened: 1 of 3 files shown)


ncordon:
___
*[build.sh, line 23 at r1](https://reviewable.io/reviews/bblfsh/scala-client/139#-LqNIgfG3mb7K4pQl07j:-LqNTb-G9A3c-44TSllX:b-dz5vd1) ([raw file](https://github.com/bblfsh/scala-client/blob/bb71d63a46cb3bc8dc088114c23ed071c51ea238/build.sh#L23)):*

How much of the savings do we get from this, vs. from excluding the irrelevant native libraries from the JAR?

I ask, because stripping the libraries might make it really hard to debug user issues.
</blockquote></details>

Without the `g++ -s` (and ignoring the `libuast` folder), the `jar` would weight 12.2MiB. With it, 7.9MiB. Last published version weighted ~10MiB (but it did not have the Windows library of course).

What about adding another compilation version for `devs`? Like `./build.sh --dev` or something like that? Or stripping only the debugging symbols (with `strip -g`) as @bzz proposed?

creachadair:
> Without the `g++ -s` (and ignoring the `libuast` folder), the `jar` would weight 12.2MiB. With it, 7.9MiB. Last published version weighted ~10MiB (but it did not have the Windows library of course).
> 
> What about adding another compilation version for `devs`? Like `./build.sh --dev` or something like that? Or stripping only the debugging symbols (with `strip -g`) as @bzz proposed?

That's probably a good idea. The case I was thinking of is in production, though, where the only evidence we get about what went wrong is the crash log. Without debug symbols in that case, we'd have nothing but offsets and function addresses. 

That said: I'm not sure we get enough value out of crash logs at this point to offset the extra cost of fetching those symbols. I was hoping the difference might be smaller, but it sounds like we get more than half the savings from that alone.

ncordon:
The way I see it is: we ask users to give reproducible examples anyway, since we cannot do much with crash logs. Therefore since I have added a command `--compile-dev` to the build script, stripping the binaries should not be very detrimental for us to debug the bugs. I have also updated the instructions of `CONTRIBUTING.md` to indicate how to compile with debug symbols

PR #140, opened 2019-10-11
Cleans JNI warnings

### LINKED ISSUE #112: v2: investigate all warnings from -Xcheck:jni
When running test for JNI using `-Xcheck:jni`:

```sh
export JVM_OPTS='-Xcheck:jni'
./sbt test
```

JVM reports number of warnings that potentially can be memory leaks or a symptoms of another stability issue like:

```
WARNING in native method: JNI call made without checking exceptions when required to from CallStaticObjectMethod
WARNING: JNI local refs: 33, exceeds capacity: 32
```

This issue is about documenting, investigating and getting rid of as much of them as possible.

ncordon:
Important warnings erased are the ones about the accumulation of local references in the `load` and `filter` methods, i.e. imagine we had a giant tree and we call `load()` over an external node. Then since methods from `Interface` in the JNI code (`Interface` is in charge of creating the nodes) were creating a lot of local references and not releasing them we could end up with an Out of Memory in the JNI side due to the accumulation of references. To exemplify this, running `testOnly org.bblfsh.client.v2.FilterNativeMethod` showed warnings about creating ~2500 local references for a quite small tree.

Left warnings when we do:

```bash
export JVM_OPTS='-Xcheck:jni'
./sbt test
```
are:

* `sbt` side warnings (before the tests actually execute). The proof for that is that if we do just `./sbt` they also arise.
* `Warning: SIGSEGV handler expected:libjvm.so+0x86be80  found:libscalauast_312939638902511669.so+0x1c7ef0` and similar: this ones disappear after we execute `test` again. I think they have more to do with `sbt`.
* 
```
WARNING: JNI local refs: 33, exceeds capacity: 32
	at java.net.NetworkInterface.getAll(Native Method)
	at java.net.NetworkInterface.getNetworkInterfaces(NetworkInterface.java:355)
	at io.netty.util.NetUtil.<clinit>(NetUtil.java:168)
	at io.netty.util.internal.MacAddressUtil.bestAvailableMac(MacAddressUtil.java:50)
	at io.netty.util.internal.MacAddressUtil.defaultMachineId(MacAddressUtil.java:138)
	at io.netty.channel.DefaultChannelId.<clinit>(DefaultChannelId.java:99)
	at io.netty.channel.AbstractChannel.newId(AbstractChannel.java:111)
	at io.netty.channel.AbstractChannel.<init>(AbstractChannel.java:83)
	at io.netty.channel.nio.AbstractNioChannel.<init>(AbstractNioChannel.java:84)
	at io.netty.channel.nio.AbstractNioByteChannel.<init>(AbstractNioByteChannel.java:66)
	at io.netty.channel.socket.nio.NioSocketChannel.<init>(NioSocketChannel.java:104)
``` 
This one is coming from `nio` library, we cannot do anything about them

*  
```
WARNING in native method: JNI call made without checking exceptions when required to from CallStaticVoidMethod
	at java.lang.Object.getClass(Native Method)
```
This one is coming from `.getClass()` method in Java. We cannot do anything about it either.

* 
```
WARNING in native method: JNI call made without checking exceptions when required to from CallStaticVoidMethod
	at java.lang.Runtime.runFinalization0(Native Method)
```
Similarly

Closes #112

---
This change is [Reviewable](https://reviewable.io/reviews/bblfsh/scala-client/140)

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -153,7 +153,7 @@ function usage() {
 }
 
 if [ "$#" -eq 0 ]; then
-    usage && exit -4
+    usage
 fi
 
 for arg in "$@"; do
--- a/build.sh
+++ b/build.sh
@@ -162,7 +162,7 @@ for arg in "$@"; do
             cleanFiles
             ;;
         "--get-dependencies")
-            getSDKFiles
+            getSDKFiles && \
             getLibuast
             ;;
         "--native")
@@ -185,4 +185,9 @@ for arg in "$@"; do
             usage
             ;;
     esac
+
+    # Avoids executing further commands if previous one failed
+    if [ $? -ne 0 ]; then
+        break;
+    fi
 done
--- a/src/main/native/jni_utils.cc
+++ b/src/main/native/jni_utils.cc
@@ -144,6 +144,7 @@ void checkJvmException(std::string msg) {
 
 jobject NewJavaObject(JNIEnv *env, const char *className, const char *initSign,
                       ...) {
+  // global ref
   jclass cls = FindClass(env, className);
   checkJvmException(std::string("failed to find a class ").append(className));
 
@@ -179,6 +180,7 @@ jfieldID FieldID(JNIEnv *env, jobject obj, const char *field,
                         .append(typeSignature)
                         .append("'"));
 
+  env->DeleteLocalRef(cls);
   return fId;
 }
 
--- a/src/main/native/jni_utils.h
+++ b/src/main/native/jni_utils.h
@@ -74,6 +74,7 @@ void checkJvmException(std::string);
 JNIEnv *getJNIEnv();
 
 // Constructs new Java object of a given className and constructor signature.
+// Returns a local reference.
 jobject NewJavaObject(JNIEnv *, const char *, const char *, ...);
 
 // Returns the field ID of the field of the given object.
@@ -86,6 +87,7 @@ jint IntField(JNIEnv *, jobject, const char *, const char *);
 
 // Reads the value of an Object field of a given object.
 // The field is specified by its name and signature.
+// Returns a local reference.
 jobject ObjectField(JNIEnv *, jobject, const char *, const char *);
 
 // Returns the method ID for a method of a given class or interface name.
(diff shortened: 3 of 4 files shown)


ncordon:
___
*[src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc, line 391 at r1](https://reviewable.io/reviews/bblfsh/scala-client/140#-Lr9eYLnBe5u3ijinHSZ:-LrA042TC7ciBuqAXOnj:b2tav27) ([raw file](https://github.com/bblfsh/scala-client/blob/853655e6f0c307091f51ceca7dd99a200a4e0418/src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc#L391)):*
<details><summary><i>Previously, kuba-- (Kuba Podgórski) wrote…</i></summary><blockquote>

For sure it's _me_, but I'm trying to understand general concept, so here what confuses me is - 
you have a function `NewJavaObject` with a comment: "Returns a local reference".
And here, the code basically says: `isLocal` if I cannot get `val->obj`, so I create a new local ref.
So, otherwise I get the _non local_?
Maybe we should rename the boolean var. or leave a _raw_ if with comment - wdyt?
</blockquote></details>

`val->obj` is a global reference (because `Node` always contains a global reference in its `jobject`). I can add a comment there clarifying that if, but yes, if `val->obj` exists what I get in `v` is not local

ncordon:
___
*[src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc, line 391 at r1](https://reviewable.io/reviews/bblfsh/scala-client/140#-Lr9eYLnBe5u3ijinHSZ:-LrA2h2t8DEfU6lGQGC-:b-765pz4) ([raw file](https://github.com/bblfsh/scala-client/blob/853655e6f0c307091f51ceca7dd99a200a4e0418/src/main/native/org_bblfsh_client_v2_libuast_Libuast.cc#L391)):*

`val->obj` is a global reference (because `Node` always contains a global reference in its `jobject`). I can add a comment there clarifying that if, but yes, if `val->obj` exists what I get in `v` is not local
</blockquote></details>

I have also renamed the variable of the `if` so it is more clear what we are doing. Just a clarification: I need that to know whether I can delete the variable afterwards or not (only if a local variable was created)

ncordon:
___
*[build.sh, line 156 at r2](https://reviewable.io/reviews/bblfsh/scala-client/140#-LrKWsJ2EhrRQTnmqCRG:-LrO5oP9BBbqU8CulFje:b499q7j) ([raw file](https://github.com/bblfsh/scala-client/blob/a95f10345dab5e860d23f8290f4a087e1b669972/build.sh#L156)):*

Since `usage` calls `exit -3`, the RHS of this conditional will never fire.
</blockquote></details>

Changed

